PR #50, opened 2016-01-27
Removed non-ascii characters in comment block.

hyoklee:
I can see  a repeated sequence of \302\240\ in comment block.


soumagne:
Joe, could you please check that? Is it only in that file or is it also present in other files?

hyoklee:
There are others (all examples files). I'll replace them globally later.

PR #52, opened 2016-01-28
add hg to hg-bulk getters

JohnPJenkins:
Convenience function for getting a bulk class from a mercury class. Especially useful when using an implicit bulk class (e.g. passing HG_BULK_NULL to HG_Init).

### DIFF
--- a/src/mercury.c
+++ b/src/mercury.c
@@ -572,6 +572,13 @@ HG_Finalize(hg_class_t *hg_class)
 }
 
 /*---------------------------------------------------------------------------*/
+hg_bulk_class_t *
+HG_Get_bulk_class(hg_class_t *hg_class)
+{
+    return HG_Core_get_bulk_class(hg_class);
+}
+
+/*---------------------------------------------------------------------------*/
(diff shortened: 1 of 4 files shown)


soumagne:
Ok I think it is fine to me to have that, I was trying to avoid to have to go this way though :) but we will see that later... could you also please move HG_Get_bulk_class()/HG_Core_get_bulk_class() right after HG_Finalize()/HG_Core_finalize() and HG_Get_bulk_context()/HG_Core_get_bulk_context() after HG_Destroy()/HG_Core_destroy() ? Just to keep the routines organized.

soumagne:
When I say I was trying to avoid to have to go this way, I actually mean that it still annoys me a little that we have to get the bulk class etc from the hg class to then use the bulk routines, but maybe when I simplify things and keep only the context in most of the routines, it will be better.

soumagne:
Ah yes that's what I meant :) thanks for understanding me :)

soumagne:
:+1:  thanks!

PR #64, opened 2016-03-04
avoid calling lookup_wait in threaded example

carns:
- The example program uses a dedicated thread to call HG_Progress,
  which means that there will be contention if one of the other threads
  calls NA_Addr_lookup_wait() (which internally will attempt to drive NA
  progress on its own)
- Use asynchronous NA_Addr_lookup() instead


soumagne:
:) OK great I think I will remove it soon then. Thanks!

PR #69, opened 2016-03-06
use short sends for req/response

### LINKED ISSUE #68: request messages are always rounded up to maximum size
This is a performance bug.  It looks like every unexpected send in Mercury is automatically sending the entire buffer of maximum unexpected size.

For CCI/TCP on loopback device, for example, this means that every request message is 8988 bytes, even when my test program is sending a request with a single integer payload.  Some overhead is expected for Mercury headers but not 8K+.  The same thing happens on the BMI transport, so it is not transport-specific. This probably makes the round trip latency significantly higher than it should be on any transport.

It looks like the problem (at some level at least) is that the NA_Msg_send_unexpected() call in mercury_core.c is using hg_handle->in_buf_size as the send size, which is the entire size of the buffer that was allocated to encode the message into, rather than the amount of the buffer that was actually used by the encoding.

carns:
This makes Mercury use short sends for both requests and responses.  Rather than sending the full buffer that was pre-allocated to the maximum unexpected size reported by the NA transport, it only sends the portion of it that holds the encoded request or response struct and HG header.

For my example program with CCI/TCP on a loopback interface, this reduces the request/response sizes in bytes from 8988/8988 to 104/28.  I think this change will probably reduce RPC latency for all transports.

Fixes #68

### DIFF
--- a/src/mercury.c
+++ b/src/mercury.c
@@ -12,6 +12,7 @@
 
 #include "mercury_hash_string.h"
 #include "mercury_proc.h"
+#include "mercury_proc_header.h"
 #include "mercury_error.h"
 
 #include <stdlib.h>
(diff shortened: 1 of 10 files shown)


soumagne:
Great! Thanks for sending that patch! That really helps me.

carns:
Thanks Jerome!

Issue #70, opened 2016-03-06
NA CCI: busy-polling with CCI/TCP


carns:
If I configure Mercury to use the CCI/TCP transport, then my server daemons consume 100% CPU (one core).  The same daemon will idle when using BMI/TCP.  This can be reproduced by running the Examples/src/example_rpc_server by itself (no clients).  It calls HG_Progress in a loop with the timeout argument set to 100.

I don't know yet if this is something in the NA layer or something in CCI itself.  Just from browsing the code, the na_cci code is not using the file descriptor poll/select method of waiting for events even when the timeout > 0, so that might be all that's needed but I'm not certain.

soumagne:
OK yes this is something that I had taken care of when implementing the BMI plugin. I don't think there is any problems in the NA layer so that must come from the CCI plugin or CCI itself, although CCI is probably doing the right thing.

soumagne:
OK we might want to include Scott on that then.

carns:
We can do this the github way by mentioning @scottatchley in the issue :)

soumagne:
@carns OK. Well it looks like you've just done it :)
@scottatchley Scott, is the problem mentioned in that issue something that you are already aware of?

scottatchley:
Adding in @gvallee

Geoffroy,

Are your TCP blocking commits all in master?

We may need to use the file descriptor in na_cci. :-)

scottatchley:
Answering my own question, his changes are in.

na_cci needs to be modified. I am trying to push a branch. I am not sure I have permissions.

carns:
@scottatchley, I've had good luck (in terms of development practice) by forking Mercury, making my changes in a branch on that fork, then opening a pull request.  It's pretty easy and it gives Jerome a chance to look over it, comment, and figure out where to land it.

soumagne:
@scottatchley I added you as an external collaborator so you should also have write access to the repo now. I'm fine with both, fork + pull request or push to topic_xxx branch in the main repo. The nice thing with the pull request is that it also triggers the travis continuous testing at the same time so we know before committing/merging that the tests pass.

scottatchley:
I cloned, committed, and pushed to my repo. I opened a pull request.

soumagne:
Hmm I just gave it a try and there is still a problem I think. If I run the mercury test server `hg_test_server -c cci`, I'm still at 100% cpu. I attached the debugger and it seems that the problem is here in na_cci_progress:
```
[...]

PR #72, opened 2016-03-07
Add blocking support to na_cci

scottatchley:
This is an untested (not even compiled) patch

### DIFF
--- a/src/na/na_cci.c
+++ b/src/na/na_cci.c
@@ -25,6 +25,7 @@
 #include <inttypes.h>
 #include <sys/queue.h>
 #include <assert.h>
+#include <poll.h>
 
 /****************/
 /* Local Macros */
@@ -150,6 +151,7 @@ struct na_cci_private_data {
 	hg_queue_t     *unexpected_op_queue;	/* Unexpected op queue */
 	hg_thread_mutex_t unexpected_op_queue_mutex;	/* Mutex */
 	char		*uri;
+	int		fd;
 };
 
 typedef union cci_msg {
@@ -570,6 +572,7 @@ na_cci_initialize(na_class_t * na_class, const struct na_info *na_info,
 	cci_endpoint_t *endpoint = NULL;
 	char *uri = NULL;
 	na_return_t ret = NA_SUCCESS;
+	int fd = 0;
 
 	/* Initialize CCI */
 	rc = cci_init(CCI_ABI_VERSION, 0, &caps);
@@ -611,13 +614,14 @@ na_cci_initialize(na_class_t * na_class, const struct na_info *na_info,
 	}
 
 	/* Create an endpoint using the requested transport */
-	rc = cci_create_endpoint(device, 0, &endpoint, NULL);
+	rc = cci_create_endpoint(device, 0, &endpoint, &fd);
 	if (rc) {
 		NA_LOG_ERROR("cci_create_endpoint() failed with %s",
 				cci_strerror(NULL, rc));
 		goto out;
 	}
 	NA_CCI_PRIVATE_DATA(na_class)->endpoint = endpoint;
+	NA_CCI_PRIVATE_DATA(na_class)->fd = fd;
 
 	rc = cci_get_opt(endpoint, CCI_OPT_ENDPT_URI, &uri);
 	if (rc) {
@@ -1918,6 +1922,13 @@ na_cci_progress(na_class_t * na_class, na_context_t * context,
 		int		rc;
 		hg_time_t	t1, t2;
 		cci_event_t	*event = NULL;
+
+		struct pollfd pfd;
+
+		pfd.fd = NA_CCI_PRIVATE_DATA(na_class)->fd;
+		pfd.events = POLLIN;
+
+		poll(&pfd, 1, ((int) remaining) * 1000);
 
 		hg_time_get_current(&t1);
 


soumagne:
Thanks for creating the pull request! I gave it a try and replied in the issue #70

src/na/na_cci.c
```diff
@@ -1919,6 +1923,13 @@ na_cci_progress(na_class_t * na_class, na_context_t * context,
 		hg_time_t	t1, t2;
 		cci_event_t	*event = NULL;
 
+		struct pollfd pfd;
+
+		pfd.fd = NA_CCI_PRIVATE_DATA(na_class)->fd;
+		pfd.events = POLLIN;
+
+		poll(&pfd, 1, ((int) remaining) * 1000);
```
carns: There is a math bug here; should be this instead:

```
poll(&pfd, 1, (int)(remaining * 1000.0));
```

The problem with the current code, is that if the value of remaining is less than 1 (remaining is a double precision floating point variable) then the cat to an integer will round it down to zero before multiplying by 1000.  We need to do the multiplication as a float before casting.

carns:
I've tested this pull request (modulo the math fix) with the CCI Verbs transport and can confirm that it does indeed reduce Mercury CPU usage from 100% down to a negligible percentage when using the na_cci plugin on CCI/Verbs.  I vote we go ahead and merge this change; if there are remaining problems within the CCI/TCP transport then they'll have to be addressed separately anyway.

carns:
Also, for informational purposes, here is a fork/branch of Mercury that has all 3 of the present Mercury pull requests (plus fixes) combined in one place.  I am using this as my working branch for now to get some mileage on the combination of na_cci changes:

https://github.com/carns/mercury/tree/topic/na-cci-create-endpoint-at-rdmaflags-blocking

That's just for informational purposes; the pull requests should be handled separately and we can't merge the create_at() changes until that is the mainstream API in CCI.

soumagne:
Hmm when running ctest, things hang with TCP so I just can't merge it now...

carns:
Ah, Ok. I'll try that early next week.

carns:
I confirmed that ctest also hangs for me when using cci/tcp in conjunction with this pull request (with or without the suggested arithmetic fix).  I have not investigated the cause yet, just sharing that I have matching behavior in that case.

soumagne:
Does the code that you were using to confirm the CPU usage with verbs also hang when using tcp? Or do you only see it when using ctest/mercury's internal tests?

src/na/na_cci.c
```diff
@@ -1919,6 +1923,13 @@ na_cci_progress(na_class_t * na_class, na_context_t * context,
...
+
 		hg_time_get_current(&t1);
```
carns: This hg_time_get_current(&t1) call needs to be moved to *before* the poll call, so that we correctly measure how much time the poll() consumed.

src/na/na_cci.c
```diff
@@ -1919,6 +1923,13 @@ na_cci_progress(na_class_t * na_class, na_context_t * context,
...
 
 		rc = cci_get_event(e, &event);
```
carns: A little further down (this patch doesn't show the line in question) there is a "remaining -= hg_time_to_double(...)" call.  There should be an if statement after that to see if remaining goes negative; if so we need to return NA_TIMEOUT right there instead of looping back around.  Otherwise the remaining variable could continue to go further and further negative.

carns:
I have it working.  There are two more corrections to the pull request above (to keep "remaining" from going negative, and to fix a timing bug), but there is also a bug in CCI/TCP that needs to be fixed for all of the tests to pass with this change.  See https://github.com/CCI/cci/issues/42.

src/na/na_cci.c
soumagne: Ok yes I had got that one.

src/na/na_cci.c
soumagne: OK I missed that one though :)

soumagne:
OK thanks for looking at it!

carns:
Just to clarify, these changes (updates to this pull request, plus fix for issue in cci repo) are enough to make ctest run correctly for Mercury with the na_cci blocking support in place.

That in turn will make Mercury run without excessive CPU usage for the cci/verbs plugin.  Unfortunately the cci/tcp plugin still burns CPU regardless of what Mercury is doing due to an orthogonal internal bug in CCI.

So ultimately this stuff is somewhat incremental (just getting pieces in place).

scottatchley:
> On Jun 6, 2016, at 4:57 PM, carns <notifications@github.com> wrote:
> 
> Just to clarify, these changes (updates to this pull request, plus fix for issue in cci repo) are enough to make ctest run correctly for Mercury with the na_cci blocking support in place.
> 
> That in turn will make Mercury run without excessive CPU usage for the cci/verbs plugin. Unfortunately the cci/tcp plugin still burns CPU regardless of what Mercury is doing due to an orthogonal internal bug in CCI.

Sorry about the slowness on our end. Geoffroy’s blocking TCP implementation used a traditional MPI technique (i.e. use a polling thread). It is on my plate to fix, but I have been distracted by our efforts to define and acquire Spider-3, our next generation center-wide file system, as well as by CORAL activities.

carns:
No problem @scottatchley.  I'm just pushing on the Mercury side of things (getting the poll fd exposed in the CCI API to interact with Mercury properly) so that everything else will fall into place once the CCI/TCP CPU usage issue is resolved.  I'm already in a scenario where I can run over CCI/verbs without busy spinning if I use the right combination of git branches.

carns:
One more more quick informational update.  If I use the combination of https://github.com/carns/mercury/tree/topic/na-cci-create-endpoint-at-rdmaflags-blocking and https://github.com/CCI/cci/tree/topic/create_endpoint_at branches of Mercury and CCI, respectively, I can confirm that the CCI SM transport also operates without busy spinning as well.

soumagne:
I was thinking I could merge that one too but looks like there is still a fix missing in CCI, I've got all your changes though so I'm ready to merge.

carns:
Ok, we still need the cci/tcp fix for blocking mode to work with that transport, right?

Should we make the na_cci blocking mode optional (selectable at runtime somehow) so that Mercury users can decide about the cpu overhead / latency tradeoff?  In the near term it would also let us get this feature upstreamed for folks that want to use it while still leaving it off by default so that nightly tests pass with cci/tcp.

soumagne:
Right.

Hmm ok yes that would work for me. I can add an option in cmake for that if you want.

carns:
That sounds good to me.  I would volunteer but I think I would do more harm than good in cmake.

I think we might eventually want it a run-time rather than compile-time option, but we can come back to that later.  For now it would be great just to have an official version of that functionality in the mainline code :-)

soumagne:
Merged in a8a60318c5a2be27dcd5bf51524e832bd48cc271 and 8cdf125d221c9342255ce1ce2b5a4613c6a2e936

Issue #48, opened 2016-01-15
address management woes


JohnPJenkins:
As discussed earlier, there are some issues with address management:
- CCI seems to "do it's own thing" with addresses. This happens for TCP (port not respected), IB, and SM (URI replaced with an entry to /tmp)
- MPI configuration (dynamic processes in particular) is cumbersome, though that's not necessarily Mercury's fault.

Getting CCI to respect user options would aid usability significantly as that's been the focus of our benchmarking so far.

More generally, initialization could do more to indicate the "right" options to provide. E.g. NA_Initialize having separate args for class, transport, and "host" where the host type is class-dependent (MPI port string or communicator/rank for MPI, hostname/port pair for TCP, filesystem entry / PID for shmem, etc.). Class-specific parser functions can be used to generate the arguments. At the moment it's unclear how exactly the input is utiltized. Additionally, clients don't need a "host" argument AFAICT, though the parsing function checks for one anyways.

ashleypittman:
We have a specific request for cci+tcp to be able to use both kernel-assigned port addresses (as it does today) but also to honour the requested port number.  I think I have a mechanism to do with the current CCI interface this but it's at best a hack.

scottatchley:
Hi Ashley,

Phil Carns has requested this as well.

I suggested a couple of possibilities, but the one I like would add cci_create_endpoint_at(cci_device_t *device, char *service, cci_os_handle_t **fd) where service would provide a transport specific hint (e.g. port number for most CTPs and PID/EndpointID for SM). In this case, the caller must specify a valid device (i.e. it cannot be NULL).

Scott

ashleypittman:
I just wrote a patch to try specifying this in device->cong_args by modifying device in the na_cci_initilize function but unfortunately this is called after the args have been read so modifying them does not have any effect.

scottatchley:
Hi Ashley,

I have pushed a branch called topic/create_endpoint_at on the CCI repo at:

https://github.com/CCI/cci/

It adds cci_create_endpoint_at() which adds a const char * service parameter:

CCI_DECLSPEC int cci_create_endpoint_at(cci_device_t * device,
                                        const char *service,
                                        int flags,
                                        cci_endpoint_t ** endpoint,
                                        cci_os_handle_t * fd);

[...]

PR #73, opened 2016-03-15
make na_cci honor NA_Iinitialize() URI

carns:
It's Ok if this isn't merged at all, or if it has some value but needs revision first.  I'm mainly opening a pull request to document the feature and start a discussion including @scottatchley @soumagne and @JohnPJenkins to see if we can find the "right" way to resolve this.

The issue, documented in part in #48, is that the Mercury API provides a URI at Iinitialization time that is intended to specify how the transport should listen for incoming requests, but this URI is ignored by na_cci in favor of the CCI config file.  This patch addresses that *crudely* by having na_cci intercept the CCI_CONFIG environment variable and redirect it to a temporary file that has been modified to use the port specified by Mercury.  It only works for TCP right now.

This is helpful for my work because it means that I can set a single, global CCI_CONFIG environment variable that works equally well for multiple servers and multiple clients in the same environment, with Mercury in charge of which ones use explicit listening ports, and what that port value is.

### DIFF
--- a/src/na/na_cci.c
+++ b/src/na/na_cci.c
@@ -24,6 +24,7 @@
 #include <stdint.h>
 #include <inttypes.h>
 #include <sys/queue.h>
+#include <unistd.h>
 #include <assert.h>
 
 /****************/
@@ -150,6 +151,7 @@ struct na_cci_private_data {
 	hg_queue_t     *unexpected_op_queue;	/* Unexpected op queue */
 	hg_thread_mutex_t unexpected_op_queue_mutex;	/* Mutex */
 	char		*uri;
+        char           *override_config_file;
 };
 
 typedef union cci_msg {
@@ -457,6 +459,9 @@ na_cci_cancel(
 	      na_op_id_t op_id
 );
 
+static char *
+override_cci_parameters(const struct na_info *na_info, na_bool_t listen);
+
 /*******************/
 /* Local Variables */
 /*******************/
@@ -511,6 +516,7 @@ na_cci_check_protocol(const char *protocol_name)
 	int		ret = 0, i = 0;
 	uint32_t	caps = 0;
 	cci_device_t	*const *devices, *device = NULL;
+        int             init_flag = 0;
 
 	/*
 	 * init CCI, get_devices, and check if a device on this transport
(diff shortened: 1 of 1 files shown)


scottatchley:
Adding @gvallee.

On Mar 15, 2016, at 2:36 PM, carns <notifications@github.com> wrote:

It's Ok if this isn't merged at all, or if it has some value but needs revision first. I'm mainly opening a pull request to document the feature and start a discussion including @scottatchley @soumagne and @JohnPJenkins to see if we can find the "right" way to resolve this.

The issue, documented in part in #48, is that the Mercury API provides a URI at Iinitialization time that is intended to specify how the transport should listen for incoming requests, but this URI is ignored by na_cci in favor of the CCI config file. This patch addresses that crudely by having na_cci intercept the CCI_CONFIG environment variable and redirect it to a temporary file that has been modified to use the port specified by Mercury. It only works for TCP right now.

This is helpful for my work because it means that I can set a single, global CCI_CONFIG environment variable that works equally well for multiple servers and multiple clients in the same environment, with Mercury in charge of which ones use explicit listening ports, and what that port value is.

You can view, comment on, or merge this pull request online at:

 https://github.com/mercury-hpc/mercury/pull/73

Commit Summary

• allow mercury to override cci parameters
File Changes

• M src/na/na_cci.c (255)
Patch Links:

• https://github.com/mercury-hpc/mercury/pull/73.patch
• https://github.com/mercury-hpc/mercury/pull/73.diff

scottatchley:
Hi all,

I am on vacation this week, but I have been checking mail occasionally.

This would require an API change or addition to CCI to support this properly. The lack of the ability to bind to a port or other identifier was debated much early and left to fix later. The main issue is that we did not want to assume all transports would use sockets for connection setup. Without assuming sockets, we were stuck with what would be the equivalent of a service/port that would be usable by each transport. As it turns out, all of the completed transports use sockets for connection handling. The only odd ball is SM which uses UNIX domain sockets and thus a path rather than a service/port.

Let me think about modifying cci_create_endpoint() or adding cci_create_endpoint_at() in order to handle transports needing a service and the SM transport needing a path.

Scott

carns:
Deprecated by #95

PR #76, opened 2016-03-19
defer RPC callback execution to HG_Trigger()

### LINKED ISSUE #61: Have HG_Trigger trigger the RPC callback
Originally we made HG_Trigger to trigger the user callbacks passed to the operations that have completed, not the actual RPC callback. The RPC callback is executed as soon as the RPC request is received, which can be source of problems.
Currently, a user callback is placed onto the trigger completion queue only when an HG operation completes, i.e. for now 2 cases: when a HG_Forward operation completes or when a HG_Response completes. The RPC callback itself gets invoked when something is received, which is not considered as an HG operation complete.

carns:
Fixes #61 

- instead of executing RPC handlers directly within HG_Progress, instead
  queue it up to be executed during HG_Trigger() call
- this avoids deadlocks in some concurrent servers that don't expect
  RPC handler work to be performed while calling Progress()

### DIFF
--- a/src/mercury_core.c
+++ b/src/mercury_core.c
@@ -63,6 +63,7 @@ struct hg_context {
     hg_class_t *hg_class;                         /* HG class */
     hg_bulk_context_t *bulk_context;              /* HG bulk context */
     hg_queue_t *completion_queue;                 /* Completion queue */
+    hg_queue_t *rpc_completion_queue;             /* Completion queue of RPCs to invoke */
     hg_thread_mutex_t completion_queue_mutex;     /* Completion queue mutex */
     hg_thread_cond_t completion_queue_cond;       /* Completion queue cond */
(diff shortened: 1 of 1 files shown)


soumagne:
I see that you added another queue, was it needed or this was more to distinguish the rpc execution callback from the other callbacks?

carns:
The latter.  I wasn't sure if there was a way to differentiate between the two cases based on just the hg_handle so I put them in separate queues.  If there is a way to tell them apart then we can simplify the patch by putting them all in the same completion queue, I think.

soumagne:
I merged it manually, not sure I followed the official github procedure on this one... :/

soumagne:
I also simplified the patch

carns:
Thanks @soumagne!  That looks much cleaner with a single completion queue.

PR #85, opened 2016-04-01
add hg_request_forward_cb convenience function

JohnPJenkins:
Simple function that should account for 90% of hg_request use cases (HG_Forward + hg_request_wait).

Also, noticed a doc error about returning -HG_XXX_ERROR, which doesn't seem to be done, and added an HG_OTHER_ERROR error type for errors outside the purview of mercury (such as mercury_util).

### DIFF
--- a/src/util/mercury_request.h
+++ b/src/util/mercury_request.h
@@ -11,6 +11,7 @@
 #ifndef MERCURY_REQUEST_H
 #define MERCURY_REQUEST_H
 
+#include <mercury_types.h>
 #include "mercury_util_config.h"
 
 /**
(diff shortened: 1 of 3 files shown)


src/util/mercury_request.h
```diff
@@ -137,6 +138,17 @@ hg_request_waitall(int count, hg_request_t *request[],  unsigned int timeout,
...
 /**
+ * A convenience callback function to pass to HG_Forward when using
+ * hg_request. This function simply interprets the argument passed into
+ * HG_Forward as an hg_request_t and calls hg_request_complete on it
+ *
+ * \return HG_SUCCESS if hg_request_complete returns HG_UTIL_SUCCESS,
+ *         HG_OTHER_ERROR otherwise
+ */
+HG_UTIL_EXPORT hg_return_t
+hg_request_forward_cb(const struct hg_cb_info *cb_info);
```
soumagne: Since the same function can be used for lookup (for now), forward, respond and bulk transfer, can you just change the name to be more generic?

JohnPJenkins: Sure! How about `hg_request_complete_cb` - a callback that calls hg_request_complete?

BTW, nice that the callback types have been merged.

soumagne: Yes that sounds good, thanks!

soumagne:
Yeah.. although now that I look at it, I wish mercury_request.h was not depending on mercury headers, since it's a separate layer. So I'm not sure if this is the right place to put that routine. I'll merge it for now and we can think later.

PR #94, opened 2016-05-04
typo in handle size bounds

JohnPJenkins:


soumagne:
I thought this was fixed a long time ago :/ I remember having seen that one multiple times before... looks like I had missed it again...

PR #95, opened 2016-05-05
update na_cci to use create_endpoint_at()

carns:
Don't merge this (yet), it depends on CCI functionality that is currently in a development branch of CCI (https://github.com/CCI/cci.git topic/create-endpoint-at)

- this makes na_cci honor the port specification in the HG listen  address
- requires the topic/create_endpoint_at branch of cci
- anything after the ":" in the HG string address is passed on to the "service" argument of cci_create_endpoint_at() if Mercury is initialized in listening mode

Thanks to @scottatchley for the CCI component of this work (i.e. the hard part)!


soumagne:
Thanks for making the fix! Yeah... I expect the check will fail because there will be a link error with the version of CCI that is used on travis

carns:
I've jumbled up this branch terribly somehow, will open another pull request later.

soumagne:
Hmm ok no problem, I will squash the commits anyway so do not worry if you have merge commits or anything weird as long as the diff is correct...

Issue #96, opened 2016-05-06
race condition between HG_Bulk_transfer and HG_Trigger


JohnPJenkins:
Haven't been able to produce a reduced test case, but it looks like there's a race condition between the NA_Put issued by a HG_Bulk_transfer and hg_bulk_trigger_entry issued by HG_Trigger, when progress/trigger and HG_Bulk_transfer are called from different threads. It's *very* unlikely to happen but my code seems to be tickling it in just the right way. The relevant valgrind output is shown below.

```
==22997== Invalid write of size 8
==22997==    at 0x67F3EE2: NA_Put (na.c:1355)
==22997==    by 0x63DE300: hg_bulk_na_put (mercury_bulk.c:221)
==22997==    by 0x63DF33B: hg_bulk_transfer_pieces (mercury_bulk.c:595)
==22997==    by 0x63DF92B: hg_bulk_transfer (mercury_bulk.c:720)
==22997==    by 0x63E16C3: HG_Bulk_transfer (mercury_bulk.c:1236)
==22997==    by 0x504AB7: margo_bulk_transfer (margo.c:576)
==22997==    by 0x48E18E: SingletonDB_HGWrapServer::PullHelper(void*, SingletonDB_HGWrapServer::pull_type) (SingletonDB_HGWrapServer.cc:128)
==22997==    by 0x48DF59: SingletonDB_HGWrapServer::PullProcRPC(void*) (SingletonDB_HGWrapServer.cc:86)
==22997==    by 0x48E692: _ZZN4sdsu11forward_rpcI24SingletonDB_HGWrapServerXadL_ZNS1_10GetMargoIDEvEEXadL_ZNS1_11PullProcRPCEPvEEEE9hg_returnS2_ENKUlS2_E_clES2_ (hg-abt-utils.hh:164)
==22997==    by 0x48E6B2: _ZZN4sdsu11forward_rpcI24SingletonDB_HGWrapServerXadL_ZNS1_10GetMargoIDEvEEXadL_ZNS1_11PullProcRPCEPvEEEE9hg_returnS2_ENUlS2_E_4_FUNES2_ (hg-abt-utils.hh:156)
==22997==    by 0x511A8F: ABTD_thread_func_wrapper (abtd_thread.c:16)
==22997==    by 0x511D30: make_fcontext (make_x86_64_sysv_elf_gas.S:64)
==22997==  Address 0xf223c10 is 0 bytes inside a block of size 8 free'd
[...]

PR #97, opened 2016-05-09
rearrange op-id store to avoid race

JohnPJenkins:
Made a quick fix to NA_Put to help mitigate the race condition in #96. It appears sufficient - tsan doesn't complain for the test case (it detects other things, though...), and I'm able to execute the program without the race corrupting glibc :).

### DIFF
--- a/src/na/na.c
+++ b/src/na/na.c
@@ -1345,14 +1345,18 @@ NA_Put(na_class_t *na_class, na_context_t *context, na_cb_t callback, void *arg,
         goto done;
     }
 
+    /* NOTE: op must be set before the put rather than after - it's possible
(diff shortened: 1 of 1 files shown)


soumagne:
Ah ok I see... when you say it detects other things do you mean related to mercury or not related?

soumagne:
Hmm ok thanks for sending that. Sorry for the delay in getting back to it, but am trying to catch up...
Also since there was an issue with the NA Put op id, I would assume there are similar issues for the other functions...

soumagne:
Yeah I saw that when skimming through the files you sent. Yes ok if you can fix the other ones and send me the patch that will be helpful and that will avoid me to context switch :)

soumagne:
Ok great thanks!

PR #99, opened 2016-05-09
reorder op_id assignment

JohnPJenkins:
Discussed in #96. Didn't touch SSM because #98. If you want, I can rebase one after you pull the other.


soumagne:
Ok no in this case I don't think you need to rebase because you didn't touch the SSM plugin so everything is good. I haven't tried yet, are most of the races you saw with tsan gone after that?

Issue #46, opened 2015-12-17
Simplify Mercury context pointer management


JohnPJenkins:
(context as in na_class/context_t and hg_class/context_t)

Currently, all four are needed - the NA pointers for address lookup, and the HG pointers for RPC registration / management. In the case where there are many "services" attaching to mercury, this strains usability. There's a number of ways in which the situation can be improved, based on the observation that every HG/NA context is associated with a single HG/NA class, and every HG class is associated with a single NA class/context:

- [x] Eliminate redundant parameters. Based on the class/context relationship, any function using the both of them can, say, drop the class pointer (HG_Init, HG_Bulk_init, HG_Create, HG_Progress, HG_Trigger, NA_Addr_lookup). This is also beneficial in the case it would be an error to, say, mismatch the class and context. Alternatively, combined with 3., you could take just a class_t when the context_t is unnecessary, and vice versa.
- [x] NA need not be exposed to the user at all - HG classes/contexts correspond to exactly one NA context / class. You could, say, have HG versions of address management functions.
- [x] Simply provide "getters" for classes and contexts. That way, I can for example just pass in an hg_context_t to a service and it can retrieve the other metadata structures as needed to perform it's work.

soumagne:
That is something we discussed in the past.
1. At some point we decided that we'd leave the API with both the class and the context to make it more uniform but we also considered storing the class pointer within the context. I'm not opposed to it.
2. I need to think a little more about that. Theoretically, an NA class could be shared between multiple HG classes, same for the HG bulk interface. Although we would need to first clean that one I think and make it use the same context as HG.
3. I'm fine with that, I think we had the HG_Get_info to retrieve things from the handle, we can probably extend that.

soumagne:
I think we're good on this one now...

PR #100, opened 2016-05-18
getters for HG class/protocol

JohnPJenkins:
This simply provides an HG API for the user to easily get at the class (not previously possible) and protocol (would have to parse HG_Addr_get_string) strings.


soumagne:
Ok I'll merge it, looks like this is another one related to bullet 3 in #46

PR #102, opened 2016-06-01
update na_cci to use new NA_MEM flag conventions

carns:
This is the na_cci change that bridges the gap between the na.h flag changes from @soumagne in Mercury and the verbs flag changes from @scottatchley in CCI.


soumagne:
Merged in f77db0d8f1acd69d3595372e94cc58c19391de37

PR #103, opened 2016-06-10
rm unnecessary APPLE check for pc generation

JohnPJenkins:
pkg-config works just fine on APPLE.

Also, does mercury really support windows??


soumagne:
Hmm ok yes I have it installed too on my mac (probably through homebrew), although does it work by default if you don't have pkg-config installed? I guess the travis check will tell me anyway...

:) yes mercury does support windows if you use MPI... Ideally we'd need just another TCP plugin to make things easier, although not a priority.

soumagne:
Ok I updated it, you may need to do a git submodule update

PR #109, opened 2016-06-16
Update mercury_error.h

jolivier23:
Users of Mercury that turn warnings to errors and use -Wall can't compile if verbose errors are disabled.  This fixes that by consuming the arguments in the macro.  The (void)0 allows a compiler to catch errors like this

if (...)
    HG_LOG_ERROR("Error Message") //missing semicolon
do_something_important();

### DIFF
--- a/src/mercury_error.h
+++ b/src/mercury_error.h
@@ -50,9 +50,9 @@
       fprintf(stdout, "\n");                                         \
   } while (0)
 #else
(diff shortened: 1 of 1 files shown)


soumagne:
This was brought in commit 46d5c25764a4ee1cbf13a1e58230193f51e7b221

PR #112, opened 2016-06-22
[WIP] address parsing

JohnPJenkins:
Here's an initial look to get an idea of what my address parsing changes would do. Strategy is to parse through the class+protocol pair, then leave the remainder for the NA layer. MPI and CCI already ignore the host name, so there was not much to do there. BMI requires reconstructing the "protocol://hostname:port", but that's a fairly simple piece of code.

Note that addr_lookup in the top-level NA layer just peeks at the class name, and passes the remainder down to the plugin. The BMI and CCI plugins then pass them directly to the libs, while MPI parses the port string starting from ';', effectively ignoring the "protocol://" portion.

TODO:

- [x] Modify tests to init clients in anon mode ("class+protocol"). Currently passes in the *server* string to initialize.
- [x] ~~Modify tests to always get the server address through file. Right now it does sometimes but not others. Will be simpler to just have one code path, then we can test things like having anon-mode servers.~~ This actually isn't the case, I misread the code.
- [x] Get rid of hard-coded tcp default protocol in tests (or just special case MPI to "dynamic") - use the NA_*_TESTING_PROTOCOL variables for this instead.
- [x] Get rid of default protocol check in test code, enforce given protocol
- [x] Turn MPI error checks (bad protocol, existence of hostname during initialization) into errors once tests are changed. They are currently warnings.
- [x] Rebase on Phil's cci_create_endpoint_at branch once it gets merged, and incorporate call into code.
- [x] Special-case address creation in test code. BMI, CCI+tcp(+verbs?) - hostname:port. CCI+sm - pid:tag (need to check format). MPI: nothing. Right now, hardcodes host string to hostname:port unless using MPI.
- [x] ~~Similarly, modify BMI to allow anon-mode servers? This might require a BMI change, if so, then just error out when no hoststring is provided at the NA layer.~~ Not worth the trouble.

Might have forgot one or two...

### DIFF
--- a/Testing/CMakeLists.txt
+++ b/Testing/CMakeLists.txt
@@ -56,18 +56,18 @@ mark_as_advanced(MERCURY_TESTING_CORESIDENT)
 # (case where the NA plugin defines multiple protocols)
 #------------------------------------------------------------------------------
 if(NA_USE_BMI)
-  set(NA_BMI_TESTING_PROTOCOL "" CACHE STRING "Protocol(s) used for testing (e.g., tcp;ib).")
+  set(NA_BMI_TESTING_PROTOCOL "tcp" CACHE STRING "Protocol(s) used for testing (e.g., tcp;ib).")
   mark_as_advanced(NA_BMI_TESTING_PROTOCOL)
 endif()
 
 #Does not really make sense for MPI (so do not add an option for it)
-#if(NA_USE_MPI)
-#  set(NA_MPI_TESTING_PROTOCOL "" CACHE STRING "Protocol(s) used for testing (e.g., tcp;ib).")
-#  mark_as_advanced(NA_MPI_TESTING_PROTOCOL)
-#endif()
+if(NA_USE_MPI)
+  set(NA_MPI_TESTING_PROTOCOL "dynamic;static" CACHE STRING "Protocol(s) used for testing (e.g., dynamic;static).")
+  mark_as_advanced(NA_MPI_TESTING_PROTOCOL)
+endif()
 
 if(NA_USE_CCI)
-  set(NA_CCI_TESTING_PROTOCOL "" CACHE STRING "Protocol(s) used for testing (e.g., tcp;ib).")
+  set(NA_CCI_TESTING_PROTOCOL "tcp;sm" CACHE STRING "Protocol(s) used for testing (e.g., tcp;sm).")
   mark_as_advanced(NA_CCI_TESTING_PROTOCOL)
 endif()
 
@@ -131,22 +131,21 @@ macro(add_mercury_test_comm test_name comm protocol opt)
     set(test_args ${test_args} --${opt})
   endif()
 
-  # Dynamic client/server test
-  add_test(NAME "mercury_${full_test_name}"
-    COMMAND $<TARGET_FILE:mercury_test_driver>
-    --server $<TARGET_FILE:hg_test_server>
-    --client $<TARGET_FILE:hg_test_${test_name}> ${test_args}
-  )
-
   # Static client/server test
-  if(${comm} STREQUAL "mpi")
-    set(static_test_name ${full_test_name}_static)
+  if(${comm} STREQUAL "mpi" AND ${protocol} STREQUAL "static")
     set(static_test_args ${test_args} --static)
-    add_test(NAME "mercury_${static_test_name}"
+    add_test(NAME "mercury_${full_test_name}"
       COMMAND ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} 1
       ${MPIEXEC_PREFLAGS} $<TARGET_FILE:hg_test_server> ${MPIEXEC_POSTFLAGS}
       ${static_test_args} : ${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MAX_NUMPROCS}
       ${MPIEXEC_PREFLAGS} $<TARGET_FILE:hg_test_${test_name}> ${static_test_args}
+    )
+  else()
+    # Dynamic client/server test
+    add_test(NAME "mercury_${full_test_name}"
+      COMMAND $<TARGET_FILE:mercury_test_driver>
+      --server $<TARGET_FILE:hg_test_server>
+      --client $<TARGET_FILE:hg_test_${test_name}> ${test_args}
     )
   endif()
 
@@ -175,8 +174,6 @@ function(add_mercury_test test_name)
       foreach(protocol ${NA_${upper_comm}_TESTING_PROTOCOL})
         add_mercury_test_comm(${test_name} ${comm} ${protocol} "")
       endforeach()
-    else()
-      add_mercury_test_comm(${test_name} ${comm} "" "")
     endif()
   endforeach()
 endfunction()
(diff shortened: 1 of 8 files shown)


soumagne:
I ljust ooked at it quickly so I may have missed one or two things but yes that looks good to me at least, thanks! Regarding the default protocol in the tests, I had also added these variables in CMakeLists.txt: NA_BMI_TESTING_PROTOCOL, NA_CCI_TESTING_PROTOCOL, etc
where you can pass a list of protocols that need to be tested per plugin, for example you can pass "tcp;sm" for CCI. Maybe that would be enough to just use that by default? and then we can just get rid of the default tcp protocol.

soumagne:
Yeah... well I guess that will be the occasion to clean it up :)

soumagne:
Nice thanks! Yes I guess we can put tcp/sm by default, should be supported everywhere and user can still remove it anyway.

soumagne:
Hmm ok, yes that seems weird...

scottatchley:
> On Jun 29, 2016, at 10:05 AM, John Jenkins <notifications@github.com> wrote:
> 
> So I tried this PR out with CCI 2.0 and it failed :(. Specifically, it hung on the mercury_rpc_cci_sm test. Works on mercury master though, which is weird since I didn't change anything in the CCI NA... Will look into this probably tomorrow.

That is odd. Right now, master and 2.0 are identical.

hyoklee:
Does id (below) mean process id? I hope you can clarify it with concrete example for CCI+SM case (and others).

> •CCI+SM expect  <id1>:<id2>  if an addr is provided.

soumagne:
Thanks, that looks good to me. Yes I would be fine with BMI just having localhost...

soumagne:
Ok thanks! let me give it a try. I think it should be documented on the website, or alternatively on the wiki for now?

soumagne:
Over there? http://mercury-hpc.github.io/documentation/#network-abstraction-layer

soumagne:
CMake Error at Testing/na/CMakeLists.txt:38 (if):
   if given arguments:

     "bmi" "STREQUAL" "mpi" "AND" "STREQUAL" "static"

soumagne:
Yes that's my fault :) some leftovers. I fixed it now, everything looks good!

soumagne:
OK wait, can you also fix the unused variables before you do that?

soumagne:
/home/jsoumagne/workspace/mercury/source/Testing/na/na_test.c: In function ‘na_test_gen_config’:
/home/jsoumagne/workspace/mercury/source/Testing/na/na_test.c:201:10: warning: unused variable ‘na_hostname’ [-Wunused-variable]
     char na_hostname[NA_TEST_MAX_ADDR_NAME];
          ^
/home/jsoumagne/workspace/mercury/source/Testing/na/na_test.c: In function ‘NA_Test_server_init’:
/home/jsoumagne/workspace/mercury/source/Testing/na/na_test.c:493:17: warning: variable ‘nret’ set but not used [-Wunused-but-set-variable]
     na_return_t nret;

/home/jsoumagne/workspace/mercury/source/src/na/na_cci.c: In function ‘na_cci_initialize’:
/home/jsoumagne/workspace/mercury/source/src/na/na_cci.c:445:11: warning: unused variable ‘string_port’ [-Wunused-variable]
     char *string_port = NULL;
           ^
/home/jsoumagne/workspace/mercury/source/src/na/na_cci.c:437:15: warning: unused parameter ‘listen’ [-Wunused-parameter]
     na_bool_t listen)

soumagne:
OK thanks! no problem ;)

soumagne:
It is going to complain because I haven't updated cci there... let me fix that

soumagne:
No the current master uses the new cci now

soumagne:
Just give me 2 seconds because I also need to clear the cache for travis

soumagne:
:) haha no it's not building from scratch every time, it builds it once and then uses the cache. Yeah unfortunately that's the only solution I found to be able to use tsan etc because ubunta packages are just buggy

soumagne:
I pushed now but it may be better to wait until the first build is done so that your pull request can use the build cache. Yes the environment is clean but you can cache dependencies so that you don't have to build them every time. In that case I specify a directory that I want to cache.
Yeah the other problem with travis is that repositories need to be white listed (which is not something I control) and unfortunately none of them have the right packages :'(

soumagne:
Yeah it usually takes about 45 minutes if we build gcc as well ;)

soumagne:
Well sorry, I forgot that the tarball has configure files already but not the repo... so have to rebuild again <sigh>

soumagne:
OK it passed this time so you can rebase now when you want

soumagne:
I merged #72 into master so just as a heads up you'll get a conflict when you rebase, you may want to squash first and then rebase... although the conflict should be easy to solve, it's only the cci_create_endpoint() stuff...

soumagne:
Yes this is normal, we are missing a fix in CCI for the tcp plugin so it will hang for now if you have NA_CCI_USE_POLL enabled, see comment in #72

soumagne:
Merged ;)

soumagne:
Looks like there are some errors with tsan
```
=================================================================
==22274==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 22 byte(s) in 1 object(s) allocated from:
    #0 0x7fb26ac0197a in malloc (/lib64/libasan.so.2+0x9897a)
    #1 0x40b538 in NA_Test_server_init /home/jsoumagne/workspace/mercury/source/Testing/na/na_test.c:442
    #2 0x405e86 in HG_Test_server_init /home/jsoumagne/workspace/mercury/source/Testing/mercury_test.c:323
    #3 0x403247 in main /home/jsoumagne/workspace/mercury/source/Testing/test_server.c:57
    #4 0x7fb26905657f in __libc_start_main (/lib64/libc.so.6+0x2057f)

Direct leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x7fb26ac0197a in malloc (/lib64/libasan.so.2+0x9897a)
    #1 0x7fb26a2e0d5d in na_bmi_addr_self /home/jsoumagne/workspace/mercury/source/src/na/na_bmi.c:866
    #2 0x7fb26a2d61f0 in NA_Addr_self /home/jsoumagne/workspace/mercury/source/src/na/na.c:629
    #3 0x40b4b6 in NA_Test_server_init /home/jsoumagne/workspace/mercury/source/Testing/na/na_test.c:436
    #4 0x405e86 in HG_Test_server_init /home/jsoumagne/workspace/mercury/source/Testing/mercury_test.c:323
    #5 0x403247 in main /home/jsoumagne/workspace/mercury/source/Testing/test_server.c:57
    #6 0x7fb26905657f in __libc_start_main (/lib64/libc.so.6+0x2057f)

SUMMARY: AddressSanitizer: 38 byte(s) leaked in 2 allocation(s).
```

soumagne:
Same with CCI

soumagne:
https://cdash.hdfgroup.org/viewTest.php?onlyfailed&buildid=20630

soumagne:
And failures on OSX now, not sure what's going on... https://cdash.hdfgroup.org/index.php?project=MERCURY

soumagne:
Closing that now since discussion was moved to #114

Issue #111, opened 2016-06-21
NA CCI: race in connection


soumagne:
Xuezhao reported the following issue (must investigate):

> One problem we observed is that sometimes mercury reports such error and cause the RPC broken:
> ```
>  NA: Error in /scratch/DAOS_OPT/src/mercury_github/src/na/na_cci.c:1678  
>  # handle_recv_unexpected(): host boro-47.boro.hpdd.intel.com, na_cci_addr->cci_addr: (nil), NA_CCI_PRIVATE_DATA(na_class)->uri: verbs://192.168.1.47:21090,  peer verbs://192.168.1.33:51218 refcnt 1
> ```
>  (the above msg is a little different with mercury master as I added some info)
> 
> It can be triggered when there are multiple client-side processes send RPC request to the server concurrently.
> 
> After some investigations, I think it is a race condition in the connection establish process:
> The server receive connect request from client, and do “cci_accept” in which will send reply to client. The client will get a “CCI_EVENT_CONNECT” event to notify the connecting done, and server will get a “CCI_EVENT_ACCEPT” event to notify the connection is established.
> 
> The problem is in some race condition, the client gets the “CCI_EVENT_CONNECT” first and then send RPC request (unexpected msg), but when the server got the unexpected msg it has not polled out the “CCI_EVENT_ACCEPT” event and firstly got another “CCI_EVENT_RECV” notification of the new unexpected msg arriving. Then when processing the unexpected msg the server treats it is invalid as the connection is not valid.
> 
> The attached simple patch is my fixing to that problem, I think that it is possibly not the best fix, but I cannot find out a better solution right now. And our testing can pass after that patch.

[0001-handle-the-possible-race-in-connection-establish.txt](https://github.com/mercury-hpc/mercury/files/324771/0001-handle-the-possible-race-in-connection-establish.txt)

scottatchley:
> On Jun 21, 2016, at 12:08 AM, Jerome Soumagne <notifications@github.com> wrote:
> 
> Xuezhao reported the following issue (must investigate):
> 
> One problem we observed is that sometimes mercury reports such error and cause the RPC broken:
> NA: Error in /scratch/DAOS_OPT/src/mercury_github/src/na/na_cci.c:1678
> 
[...]

Issue #107, opened 2016-06-15
segfault on bad address strings


carns:
If you pass an improperly formatted string into HG_Init() then it will cause a segmentation fault rather than returning an error.  This can be tricky to observe/debug in a job batch script environment.

The example below shows the segfault backtrace from calling HG_Init() with "blahblah" as the string.  It probably just needs a couple of safety checks on the string parsing.

```
#0  strlen () at ../sysdeps/x86_64/strlen.S:106
#1  0x00007ffff701fa4e in __GI___strdup (s=0x0) at strdup.c:41
#2  0x0000000000417662 in na_info_parse (info_string=0x7fffffffe105 "blahblah", 
    na_info_ptr=0x7fffffffdb18)
    at /home/pcarns/working/src/sds/mercury/src/na/na.c:197
#3  0x0000000000417aa1 in NA_Initialize (info_string=0x7fffffffe105 "blahblah", 
    listen=1 '\001') at /home/pcarns/working/src/sds/mercury/src/na/na.c:297
#4  0x0000000000410716 in hg_core_init (
    na_info_string=0x7fffffffe105 "blahblah", na_listen=1 '\001', 
    na_init_class=0x0, na_init_context=0x0)
    at /home/pcarns/working/src/sds/mercury/src/mercury_core.c:936
#5  0x000000000041400b in HG_Core_init (
    na_info_string=0x7fffffffe105 "blahblah", na_listen=1 '\001')
    at /home/pcarns/working/src/sds/mercury/src/mercury_core.c:2195
#6  0x00000000004082fc in HG_Init (na_info_string=0x7fffffffe105 "blahblah", 
    na_listen=1 '\001')
    at /home/pcarns/working/src/sds/mercury/src/mercury.c:576
#7  0x0000000000403411 in main (argc=2, argv=0x7fffffffdd18)
    at ../examples/server.c:53
```

JohnPJenkins:
I'm going to take a look at the address parsing code today, try and clean up the error cases as well as relax the parsing to allow for anonymous addresses. Should clear out a bunch of bugs.

soumagne:
Ok thanks!

soumagne:
Is this one fix then?

JohnPJenkins:
Possibly :). I'm patching up mercury-benchmarks now - I'll try it out with some "bad" addresses when I'm finished.

JohnPJenkins:
[...]

PR #114, opened 2016-07-01
tests leak fix

JohnPJenkins:
Fixes leaks introduced by #112

### DIFF
--- a/Testing/na/na_test.c
+++ b/Testing/na/na_test.c
@@ -447,6 +447,9 @@ NA_Test_server_init(int argc, char *argv[], na_bool_t print_ready,
 
     na_test_set_config(addr_string);
 
+    free(addr_string);
+    NA_Addr_free(na_class, self_addr);
+
     /* As many entries in addr table as number of server ranks */
     if (addr_table_size) *addr_table_size = na_addr_table_size;
 
--- a/Testing/na/na_test.c
+++ b/Testing/na/na_test.c
@@ -447,7 +447,7 @@ NA_Test_server_init(int argc, char *argv[], na_bool_t print_ready,
 
     na_test_set_config(addr_string);
 
-    free(addr_string);
+    free(addr_string); 
     NA_Addr_free(na_class, self_addr);
 
     /* As many entries in addr table as number of server ranks */
--- a/Testing/na/test_client.c
+++ b/Testing/na/test_client.c
@@ -211,6 +211,7 @@ main(int argc, char *argv[])
     na_return_t na_ret;
     unsigned int i;
 
+    printf("ZZZ na client\n");
     /* Initialize the interface */
     params.na_class = NA_Test_client_init(argc, argv, server_name,
             NA_TEST_MAX_ADDR_NAME, NULL);
--- a/Testing/na/test_server.c
+++ b/Testing/na/test_server.c
@@ -254,6 +254,7 @@ main(int argc, char *argv[])
     struct na_test_params params;
     na_return_t na_ret;
 
+    printf("ZZZ na server\n");
     /* Initialize the interface */
     params.na_class = NA_Test_server_init(argc, argv, NA_TRUE, NULL, NULL,
         &number_of_peers);
--- a/Testing/test_server.c
+++ b/Testing/test_server.c
@@ -46,6 +46,7 @@ hg_progress_thread(void *arg)
 int
 main(int argc, char *argv[])
 {
+    printf("ZZZ hg server\n");
     hg_class_t *hg_class = NULL;
     hg_context_t *context = NULL;
     unsigned int number_of_peers;


soumagne:
Hmm does not seem to be fixed: https://cdash.hdfgroup.org/viewDynamicAnalysis.php?buildid=20646

soumagne:
Yeah the check only means that the configure/build/test were run sucessfully, you still need to look at the dashboard to check that test results are passing or not. I can probably make it more strict so that it actually fails every time a test does not pass.

soumagne:
Just check the travis build number in the build name: travis-ci-linux-x64-gcc-debug-addresssanitizer-231

soumagne:
My problem is that if I make it more strict, it will fail for everything, even for random failures like we currently have with the CCI tcp plugin... so I would rather keep it more permissive for now.

soumagne:
I just tried on my laptop and I don't see the leaks anymore with tsan, I'm not sure why it shows up with travis still...

soumagne:
Well we don't test sm on travis so that's normal. Also Cdash only collects the results that the travis build submits, it's not doing any test itself.

soumagne:
I'm using 5.3.1 here and it's gone too for me. I suspect travis is not fetching the pull request correctly...

soumagne:
Can you try to push a fake commit to that pull request to trigger the build again?

soumagne:
Yes there were some failures, I forgot now the actual error though I would have to try again

soumagne:
Hmm that's really bizarre, same failures again in the dashboard after your fake commit...

soumagne:
Yes I tried those already, it's buggy, and same for that https://bugs.launchpad.net/ubuntu/+source/gcc-5/+bug/1413474

soumagne:
I don't think it's a compiler issue... ok I have another idea :) can you try to push a commit that prints a line in the test so that we can see if we're testing the right version?

soumagne:
Yes there is an issue with the na_simple_cci* tests, I saw that too, I haven't figured out yet why it fails, that's the only ones that fail.

hyoklee:
Awesome, @JohnPJenkins!

soumagne:
As I suspected, I don't see the string, so that means it's indeed not testing the right version... funny

soumagne:
Maybe they just broke something in the travis code

soumagne:
Could that be related to #111 ?

soumagne:
`./src/plugins/ctp/sm/ctp_sm_api.c:1674:	node = tsearch(sconn, &sep->conns, sm_compare_conns);`

soumagne:
We can pass options to Ctest using CTEST_MEMORYCHECK_SANITIZER_OPTIONS, see for example in Testing/script/travis_script.cmake line 88, so I think we could do that

soumagne:
Well but that's another topic, I would like to understand first why travis is not testing the right revision in the pull requests...

soumagne:
Here is what travis does:
```
0.60s$ git clone --depth=50 https://github.com/mercury-hpc/mercury.git mercury-hpc/mercury

Cloning into 'mercury-hpc/mercury'...

remote: Counting objects: 427, done.

remote: Compressing objects: 100% (280/280), done.

remote: Total 427 (delta 246), reused 247 (delta 144), pack-reused 0

Receiving objects: 100% (427/427), 270.00 KiB | 0 bytes/s, done.

Resolving deltas: 100% (246/246), done.

Checking connectivity... done.

$ cd mercury-hpc/mercury

0.30s$ git fetch origin +refs/pull/114/merge:

remote: Counting objects: 18, done.

remote: Compressing objects: 100% (17/17), done.

remote: Total 18 (delta 14), reused 4 (delta 1), pack-reused 0

Unpacking objects: 100% (18/18), done.

From https://github.com/mercury-hpc/mercury

 * branch            refs/pull/114/merge -> FETCH_HEAD

$ git checkout -qf FETCH_HEAD
```

soumagne:
OK! I'll post here if I find something. For now what I can say is that the git command is correct and gets your commit with the string but the string is not in log. You can get the log by clicking on raw log and look for the string which should appear just before the listen printf.

soumagne:
osx fix in c98da537119d4732190e477de7fcfdb1558d235c and pull request ctest fix in 5f9aa818bf1e75bd53937ef8d5fd585b7f6aba49

soumagne:
Yes I fixed the issue on OS X (segfault with locator in na_info_parse) and the pull request problem where ctest was actually doing a checkout of another revision so it's not doing that anymore. I haven't fixed the na_simple_* test though but that should be a separate issue.

soumagne:
Oh and before we change topic and forget, let me know when you have a look at #107

soumagne:
Oh no problem, just meant when you have time, that's really minor anyway, thanks!

scottatchley:
Hi John,

I just pushed a commit to CCI that fixes several leaks in the SM connection path. Can you recheck the below or are you filtering out CCI internal leak messages?

Scott

soumagne:
Yes it reports thread sanitizer and address sanitizer results. https://cdash.hdfgroup.org/index.php?project=MERCURY The interface should be updated and cleaner when we update to the latest version of cdash but the IT person has not gotten to it yet.

We are not filtering out CCI leak messages.

soumagne:
Also travis does not test sm for now, I have to get back to it as they were some issues when I enabled it, I can try again though.

scottatchley:
Thanks for the check!

Are there any outstanding issues with the sm transport? Blocking mode? Race conditions? Other?


________________________________

carns:
Blocking mode on the cci/sm transport works great for me!  Not even a hiccup.

scottatchley:
Ok, thanks for the confirmation. I will continue to focus on verbs. I will probably write a CCI native test that spawns multiple threads on both a client and server and repeatedly connect, communicate, disconnect, and repeat.

soumagne:
@scottatchley , you might want to look at that: https://cdash.hdfgroup.org/viewDynamicAnalysisFile.php?id=46913
I enabled the sm transport in travis and am getting some errors.

carns:
FYI, the RMA options specifically will fail when using CCI/SM on some linux distros if you don't set /proc/sys/kernel/yama/ptrace_scope to 0 (as root).  This loosens permission restrictions on that system call to allow two pids owned by the same users to access each other's memory.

If CCI/SM gets a predictable errno out of the cma system calls in this case then maybe it could print a verbose error message suggesting that fix?  A few of us have bumped into it at different times.

Not sure if that's what's going on, but it might be worth checking in that docker image.

soumagne:
Hmm ok thanks I just checked and it's indeed set to 1 so I guess that's the problem. I can't change that value though as this image does not allow me to be root unless I use another image which requires other changes and increases the boot time.

soumagne:
1 - restricted ptrace: a process must have a predefined relationship
    with the inferior it wants to call PTRACE_ATTACH on. By default,
    this relationship is that of only its descendants when the above
    classic criteria is also met. To change the relationship, an
    inferior can call prctl(PR_SET_PTRACER, debugger, ...) to declare
    an allowed debugger PID to call PTRACE_ATTACH on the inferior.
    Using PTRACE_TRACEME is unchanged.

I guess we could also call prctl in that case...

scottatchley:
You beat me to this.

This is only required if using CMA. If using the mmap bounce buffers, it is not needed.

Perhaps we can add the prctl() call if CMA is enabled. I am hesitant to do that in a library without warning the application user that this security feature has been disabled.

Thoughts/comments?

scottatchley:
I just reread the process_vm_[write|read]v() manpage. If it fails with EPERM, I have added the following message:

ctp_sm_rma: process_vm_[write|read]v() failed because it cannot trace another process. See https://www.kernel.org/doc/Documentation/security/Yama.txt for more details.

soumagne:
Ah ok then for the time being I could just simply try to have CMA disabled on the travis images, I also understand now why only the tests that call NA_Put/NA_Get were failing.

Hmmm yes that's true, although that's probably not as worse as setting ptrace_scope to 0 as root... Just googled it to see if others had the same issue and found that discussion: https://www.open-mpi.org/community/lists/users/2015/02/26391.php ... looks like their decision was to enable CMA if no yama and no PR_SET_PTRACER.

Maybe the best solution is to keep the verbose error message or have an option force CMA I guess which would call prctl() and throw a warning...

scottatchley:
CCI has --with-cma to enable CMA. Perhaps I should change the default to --without-cma and include a message in the configure --help output that enabling CMA will force the ability to trace? I could put more detail in the sm README.

soumagne:
I added `prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0)` if sm is used in the mercury testing suite and everything is fine now with CMA.

Just been thinking a little more and if enabling CMA in CCI changes permissions under the hood then it may do too much in some cases if let's say somebody only wants to have CMA enabled between parent and child processes and not any process. I think the best option is to just keep the verbose error message and let the app decide :)

carns:
For my 2c, I like the idea of keeping cma on by default and just doing what we can via docs/error messages to tell people how to either turn it off or set permissions.

I'm worried about the failure mode of accidentally compiling without CMA support and then being confused by the resulting performance :)

scottatchley:
Ok, I will leave configure alone. If it detects CMA, it will compile it in. If compiled in and it fails, it will print the warning once when the first RMA fails with EPERM.

PR #118, opened 2016-07-10
add note about patching BMI for OSX

carns:
The BMI repository referred to in the Mercury documentation will not build on OSX without first applying the patch described in this README update, provided by Rob Ross.


soumagne:
Actually I had that patch in mercury in Testing/script, I guess I could remove it now as well

carns:
Ah, I didn't realize that the patch was there as well.  Probably for the best if it lives in the BMI repo in the long run, though.  Thanks for merging the doc update.

PR #124, opened 2016-08-05
reduced memory allocation

chuckcranor:
I looked at the mallocs in mercury in the critical path.  For one noop operation a 
mercury client does 14 memory allocations.  On the server it does 12 memory 
allocations per bake noop operation.

I looked at the mercury code and spent some time restructuring
it to reduce the number of memory allocations in the critical path,
assuming that the client will cache the hg_handle it uses to communiate
with the server.

Changes I made include:
       
      - removed mallocs associated with the hg_context pending
        and processing lists by converting them into a <sys/queue.h>
        LIST.

      - removed malloc associated with the hg_forward_cb_info
        structure and carry the info in the handle.  I had to
        add a new internal forwarding callback type to make this
        work with the current layering in the code.

      - removed memory allocation of hg_completion_entry by embedding it
        in other structures.  the completion queue becomes a simple TAILQ.

      - removed hg_queue allocate of na completion queue by converting
        it to a TAILQ

      - moved allocation of na_cb_completion_data and na_cb_info
        into the op_id data structures (so when you allocate an
        op_id, you get the na_cb_completion_data and na_cb_info
        with them)

     - add framework to allow higher level code to preallocate na send
       and recv op_id's at handle creation time.   use framework to
       allow high level code to pass preallocated NA op_id's down into
       the na layer at message send/recv time  (feature is optional)
       this involves two new NA calls: NA_Prealloc_op_id()
       and NA_Prealloc_op_id_free().

       currently applies only to expected/unexpected send/recv since
       that's that bake-bulk noop uses.   if you cache your hg_handles,
       the cache will include the preallocated NA structures.

     - added support for preallocated NA op_id's to BMI and CCI
       (not currently compiling MPI, so didn't do that one)

those changes combined with caching the hg_handle on the client
side reduce the number of critical path memory allocations for noop 
to zero.

[here's a detailed list of current mercury memory allocations on the
 critical path for a noop call]

     For the client side there are 14 memory allocations per noop:

(allocate handle)
  [1] malloc 184 bytes for hg_handle
    [2] posix_memalign 252 bytes (align=4096) for handle->in_buf
    [3] posix_memalign 252 bytes (align=4096) for handle->out_buf
(forward request)
  [4] malloc 32 bytes for hg_forward_cb_info   (HG_Forward)
    [5] calloc 128 bytes for a na_cci_op_id_t  (recv_expected: for reply)
    [6] calloc 128 bytes for a na_cci_op_id_t  (send_unexpect: for request)
(send completes)
    [7] malloc 40 bytes for na_cb_info struct
    [8] malloc 32 bytes for na_cb_completion
    [9] malloc 24 bytes for hg_queue_entry  (na_private_context->completion)
         /* trigger callback will discard, i think
            since na_completed_count == 1 */
(reply arrives)
    [10] malloc 40 bytes for na_cb_info struct
    [11] malloc 32 bytes for na_cb_completion
    [12] malloc 24 bytes for hg_queue_entry  (na_private_context->completion)
          /* trigger callback has na_completed_count == 2, can pass up */
(rpc now complete)
    [13] malloc 16 bytes for hg_completion_entry struct
    [14] malloc 24 bytes for hg_queue_entry struct
(the callback will now be triggered)

the server side has 12 memory allocations (i'm not counting any of
the init stuff that happens during server startup):

(server gets unexpected recv at NA layer and queues it at NA layer)
    [1] malloc 40 bytes for a na_cb_info struct
    [2] malloc 32 bytes for a na_cb_completion
    [3] malloc 24 bytes for a hg_queue_entry (for push)
(server HG callback triggered [hg_core_recv_input_cb()])
    [4] malloc 24 bytes for a hg_list_entry_t  (for "processing" list)
    [5] malloc 16 bytes for a hg_completion_entry
    [6] malloc 24 bytes for a hg_queue_entry   (put on HG completion queue)
(server collects req and runs noop)
(server sends reply with HG_Respond() on handle)
    [7] calloc 128 bytes for a na_cci_op_id_t    (sending reply)
(server's reply send completes)
    [8] malloc 40 bytes for na_cb_info struct
    [9] malloc 32 bytes for na_cb_completion
    [10] malloc 24 bytes for hg_queue_entry  (na_private_context->completion)
(rpc now complete-queue completion for server)
    [11] malloc 16 bytes for hg_completion_entry struct
    [12] malloc 24 bytes for hg_queue_entry struct

### DIFF
--- a/Testing/na/test_cancel_client.c
+++ b/Testing/na/test_cancel_client.c
@@ -98,7 +98,7 @@ msg_unexpected_send_cb(const struct na_cb_info *callback_info)
         sprintf(params->send_buf, "Hello again Server!");
         ret = NA_Msg_send_unexpected(params->na_class, params->context,
             NULL, NULL, params->send_buf, params->send_buf_len, params->server_addr,
-            NA_TEST_SEND_TAG, NA_OP_ID_IGNORE);
+            NA_TEST_SEND_TAG, NULL, NA_OP_ID_IGNORE);
         if (ret != NA_SUCCESS) {
             fprintf(stderr, "Could not start send of unexpected message\n");
         }
@@ -177,7 +177,7 @@ msg_expected_send_cb(const struct na_cb_info *callback_info)
         printf("Sending again local memory handle...\n");
         ret = NA_Msg_send_expected(params->na_class, params->context,
             NULL, NULL, params->send_buf, params->send_buf_len, params->server_addr,
-            NA_TEST_BULK_TAG, NA_OP_ID_IGNORE);
+            NA_TEST_BULK_TAG, NULL, NA_OP_ID_IGNORE);
         if (ret != NA_SUCCESS) {
             fprintf(stderr, "Could not start send of memory handle\n");
             return ret;
@@ -208,7 +208,7 @@ test_send(struct na_test_params *params)
     /* Preposting response */
     na_ret = NA_Msg_recv_expected(params->na_class, params->context,
         msg_expected_recv_cb, params, params->recv_buf, params->recv_buf_len,
-        params->server_addr, send_tag + 1, &op_id);
+        params->server_addr, send_tag + 1, NULL, &op_id);
     if (na_ret != NA_SUCCESS) {
         fprintf(stderr, "Could not prepost recv of expected message\n");
         return EXIT_FAILURE;
@@ -223,7 +223,7 @@ test_send(struct na_test_params *params)
 
     na_ret = NA_Msg_recv_expected(params->na_class, params->context,
         msg_expected_recv_cb, params, params->recv_buf, params->recv_buf_len,
-        params->server_addr, send_tag + 1, NA_OP_ID_IGNORE);
+        params->server_addr, send_tag + 1, NULL, NA_OP_ID_IGNORE);
     if (na_ret != NA_SUCCESS) {
         fprintf(stderr, "Could not prepost recv of expected message\n");
         return EXIT_FAILURE;
@@ -232,7 +232,7 @@ test_send(struct na_test_params *params)
     /* Try to cancel unexpected send */
     na_ret = NA_Msg_send_unexpected(params->na_class, params->context,
         msg_unexpected_send_cb, params, params->send_buf, params->send_buf_len,
-        params->server_addr, send_tag, &op_id);
+        params->server_addr, send_tag, NULL, &op_id);
     if (na_ret != NA_SUCCESS) {
         fprintf(stderr, "Could not start send of unexpected message\n");
         return EXIT_FAILURE;
@@ -284,7 +284,7 @@ test_bulk(struct na_test_params *params)
     na_ret = NA_Msg_recv_expected(params->na_class, params->context,
         ack_expected_recv_cb, params, params->recv_buf, params->recv_buf_len,
         params->server_addr, ack_tag,
-        NA_OP_ID_IGNORE);
+        NULL, NA_OP_ID_IGNORE);
     if (na_ret != NA_SUCCESS) {
         fprintf(stderr, "Could not start receive of acknowledgment\n");
         return EXIT_FAILURE;
@@ -294,7 +294,7 @@ test_bulk(struct na_test_params *params)
     printf("Sending local memory handle...\n");
     na_ret = NA_Msg_send_expected(params->na_class, params->context,
         msg_expected_send_cb, params, params->send_buf, params->send_buf_len,
-        params->server_addr, bulk_tag, &op_id);
+        params->server_addr, bulk_tag, NULL, &op_id);
     if (na_ret != NA_SUCCESS) {
         fprintf(stderr, "Could not start send of memory handle\n");
         return EXIT_FAILURE;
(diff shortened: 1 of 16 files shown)


soumagne:
I haven't gone through all the changes yet but from what I just saw that looks quite good to me (even if some tests are failing for now :) ). A couple of things I'd like to keep though:
 - I'd like us not to use the TAILQ/LIST etc directly but keep using the mercury_queue / mercury_list stuff, maybe that just means improving it to provide macros that map to TAILQ/LIST so that we skip the allocation?
 - For the NA op ID passed to the NA calls, maybe we can just keep one parameter? instead of having 2 op_ids parameters

carns:
FWIW, I've become a big fan of utlist (and the other ut* companion data structures) for this sort of thing recently:
https://github.com/troydhanson/uthash/tree/master/src
The utlist.h file is standalone and can just be copied into a project, and it has a very permissive license.

carns:
Here is a better overview of utlist: https://troydhanson.github.io/uthash/utlist.html

soumagne:
Ah interesting actually I saw it recently in the MDHIM tree as well ;)

chuckcranor:
I'm still processing, but I looked over the test output and discovered that "ctest" under the configuration of mercury I've been using does 40 tests, while the one on travis does 105.   

It turns out that none of the 40 tests that my build was doing ran the hg_core_forward_self() code path, and there was one line that needed to be updated on that code path.  I was able to manually trigger the bug without using "ctest" by running "hg_test_rpc --comm bmi --protocol tcp --self" ...

chuckcranor:
Hmm, I found another malloc() on the server side on the critical path.   When you start a server (listening==true), it creates/malloc's and posts HG_MAX_UNEXPECTED_RECV (defined to 256) hg_handles to the pending list (in the hg_context).   That's done by the function hg_core_listen().  

The handles on the pending list get consumed as new requests come in to the server.  The processing function (HG_Core_progress()) continues to call hg_core_listen() if you are listening for requests. The hg_core_listen() checks to see if the pending list of handles in the context is not empty.   If the list is not empty (the normal case), hg_core_listen() doesn't do anything -- it just returns.

But if the list of pending handles in the context drops to zero (list is empty), then it allocates another set of HG_MAX_UNEXPECTED_RECV (256) handles for future requests.

So the behavior of the server is to process HG_MAX_UNEXPECTED_RECV (256) requests without allocating memory (it uses up the pending list).   Then on the next request it will do a burst of allocs while allocating another set of 256 handles for recv unexpected.  When I was doing detailed testing, I didn't run enough requests to trip on the 256 boundary to see this.

I think a way to handle this is to modify the code so that when the server is finished with a handle from the pending list that was allocated by hg_core_listen(), it should add it back to the pending list for reuse/recycle rather than freeing it (and using up the pending list).   That way we'd only do the malloc in hg_core_listen() if we are getting so many unexpected receives that we need more handles in the pending list pool than we currently have.

It seems like the key place here is hg_core_trigger_entry().  Currently for the "Execute user callback" case it unconditionally frees the handle after running the callback.   We need some way to tell if the current hg_handle is from the pending list, and if so we want to reset it and put it back on the pending list so it can be reused by the server (rather than freeing it and making the server malloc more down the line).

This only impacts the NA code path... for the "self" code path the handle from HG_Forward is directly used in the reply (so it doesn't use pending list handles in that case, I think).

chuckcranor:
I had thought about passing the preallocated NA op_id into the NA send/recv calls using the op_id that was already there (e.g. changing it from an OUT parameter into an IN/OUT parameter), but thought it would be easier/safer to do it as an extra arg to start with and then come back later to look at merging them together once things were stable.

My intent in using the <sys/queue.h> macros was to get around the mallocs in the mercury_queue/mercury_list stuff, and sys/queue.h was already in the build so it fit with the code already there.  utlist would be ok too, i think.

soumagne:
Re the 256 unexpected recv allocation, I agree, that's something we should definitely optimize and re-use the handles, that's also part of the fixes that I'd like to go in so that the NA plugins do the initial allocations etc.

Re the op ids and the macros stuff, yes no worries, I think that is already quite good that you went through that, we'll just clean it up and fix the existing the util stuff.

chuckcranor:
I just pushed an update to let the server cache the 256 unexpected recv handles and recycle them when done.  It works with the NA preallocation change I did earlier (so those structures get allocated at the same time as the handle).

On a somewhat related note, the address sanitizer output showed an issue at shutdown with the BMI na I'm looking at (a preallocated BMI op_id is getting freed at shutdown, and then the memory is referenced after the free).   To address this I think I need to add a reference counter to the BMI op_id (the CCI op_id already has a ref counter).   The reference count should be 2 at the shutdown (one for the preallocation and one for the NA that is currently using the op_id).  Dropping the preallocation should drop the reference count to 1 (instead of freeing), then when the NA is done it will drop the final reference and free.  At least that what I'm currently thinking (have tried to add it yet).

soumagne:
Thanks! Hopefully I can start getting through your changes soon. Also please send me the CLA stuff when you get that done.

I see, yes from what you're describing that seems right to me. Should be fine to add a refcount I think.

chuckcranor:
I'm working on getting approvals for signing the CLA.   Will advise.

soumagne:
Just to let you know, I'm now working on fixing mercury queue so that it uses atomics as well (and we can also get rid of all the mutexes for insertion/retrieval). So we'll definitely not use TAILQ or utlist directly in the code.

soumagne:
Not that I know except what exists through opa_queue etc.

soumagne:
I created a separate issue https://github.com/mercury-hpc/mercury/issues/127 for that

soumagne:
Had to rework a bunch of things but I've got everything clean and checked in now. There are a few more things (e.g., proc create needs to be moved to HG_Create) but I'm also fixing it. Thanks!

chuckcranor:
nice, thanks for getting the fixes in!

PR #128, opened 2016-08-10
match proc function signatures for provided types

JohnPJenkins:
The following function signature (and it's peers):
hg_return_t hg_proc_hg_uint8_t(hg_proc_t proc, hg_uint8_t *data);

Doesn't match the expected function pointer type:
hg_return_t (*hg_proc_cb_t)(hg_proc_t proc, void *data);

This trips up C++ compilers in particular - they won't convert the function
pointer type. I believe C99 complains as well, but I haven't tried.

In any case, it's a simple enough fix - type-pun inputs to `void*`'s and cast back in the function bodies.


soumagne:
I compile with pedantic and C11 and it doesn't complain :) anyway yes that should match the callback. I will merge it. Thanks!

PR #133, opened 2016-08-18
hg_proc_memcpy_decode_in_place()

carns:
A variation on the existing hg_proc_memcopy() routine that just sets a pointer to refer to a memory address in the original message, rather than copying out to an external buffer.  This can be used to avoid one extra malloc/free pair when eagerly packing raw data payloads into RPC requests or responses.


soumagne:
Maybe I'm missing something but how is it different from the hg_proc_save_ptr() / hg_proc_restore_ptr() routines that were also added?

carns:
It's different in that I didn't know about the save/restore_ptr() functions :)

I'll have a look at those and report back.  There might not be any reason to merge this pull request.

soumagne:
I think I modified those recently when I fixed the bulk handle encoding and I may have forgotten to tell anyone about it :)
@liuxuezhao you may also want to look at the save/restore_ptr functions instead of using HG_Core directly...

carns:
Yep, I'm closing this pull request.  I tested the save/restore_ptr proc functions just now and they work great.

soumagne:
OK great thanks!

PR #134, opened 2016-08-18
remove unecessary poll() call in na_cci if remaining time is zero

carns:
Reduce latency by skipping the poll() system call if the na_cci progress loop has no timeout, or if timeout has been exhausted in previous iteration.

### DIFF
--- a/src/na/na_cci.c
+++ b/src/na/na_cci.c
@@ -1896,10 +1896,13 @@ na_cci_progress(na_class_t * na_class, na_context_t * context,
         hg_time_get_current(&t1);
 
 #ifdef NA_CCI_HAS_POLL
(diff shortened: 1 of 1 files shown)


soumagne:
Just curious, does this reduce latency a lot?

carns:
It helps a little bit, but nothing earth shattering.

One thing I am curious about is if there is any significant difference between busy-spinning at the Mercury API level (by calling progress with 0 timeout) vs. busy-spinning at the na_cci layer (by disabling the USE_POLL feature).  It may help the former, which would be helpful in having the ability to control polling behavior a little further up the stack.  I haven't confirmed, though.

soumagne:
OK. Hmm yes not sure either, let's see once I add polling at the NA layer...

Issue #142, opened 2016-10-12
user facing hooks for max rpc sizes


JohnPJenkins:
We're currently using a hand-rolled eager transfer mode as a separate RPC type as an alternative to the built-in mercury eager bulk mode. We do this to prevent calls to HG_Bulk_create. While this works, we can only currently guess (or read the source code :)) of the sizes at the NA layer for which this approach saves us work - as background for others, RPCs that exceed a size threshold (NA_Msg_max_unexpected_size) use a bulk handle under the covers to do a staged transfer.

Since the user no longer sees NA functions, the semantic that matters is knowing what the max size of any particular RPC is in order to make buffering decisions. Hence, I propose the API:

```C
/**
 * Returns the threshold size by which RPC arguments can be transferred in the most
 * efficient manner for the underlying network abstraction.
 *
 * \param [in] hg_class the class to check
 * \return the size threshold, or 0 if an invalid class was passed in.
 */
hg_size_t HG_Get_rpc_threshold_size(hg_class_t *hg_class);
```

I can put together a PR if we agree on the API. The implementation is trivial - it just calls `NA_Msg_get_max_unexpected_size` and subtracts the mercury header size.

JohnPJenkins:
As soon as I hit send, I thought of the dual case - RPC return arguments. So really, there'd be two functions:

```C
/** ... */
hg_size_t HG_Get_rpc_in_threshold_size(hg_class_t *hg_class);
/** ... */
hg_size_t HG_Get_rpc_out_threshold_size(hg_class_t *hg_class);
```

"in" works along the `HG_Forward`/`HG_Get_input` path, "out" works along the `HG_Respond`/`HG_Get_output` path.

scottatchley:
> On Oct 12, 2016, at 10:13 AM, John Jenkins <notifications@github.com> wrote:
> 
> We're currently using a hand-rolled eager transfer mode as a separate RPC type as an alternative to the built-in mercury eager bulk mode. We do this to prevent calls to HG_Bulk_create. While this works, we can only currently guess (or read the source code :)) of the sizes at the NA layer for which this approach saves us work - as background for others, RPCs that exceed a size threshold (NA_Msg_max_unexpected_size) use a bulk handle under the covers to do a staged transfer.
> 
> Since the user no longer sees NA functions, the semantic that matters is knowing what the max size of any particular RPC is in order to make buffering decisions. Hence, I propose the API:
> 
> /**
[...]

PR #143, opened 2016-10-12
[WIP] size getters for eager mode RPC arg xfers

JohnPJenkins:
See #142 for discussion. This initial implementation doesn't account for alignment/etc., unclear whether it's necessary.

### DIFF
--- a/src/mercury.c
+++ b/src/mercury.c
@@ -761,6 +761,28 @@ HG_Class_get_protocol(const hg_class_t *hg_class)
 }
 
 /*---------------------------------------------------------------------------*/
+hg_size_t
+HG_Class_get_input_eager_size(const hg_class_t *hg_class)
+{
+    hg_size_t ret = HG_Core_class_get_input_eager_size(hg_class);
+    if (ret > HG_CHECKSUM_MAX_SIZE)
+        return ret - HG_CHECKSUM_MAX_SIZE;
+    else
+        return 0;
+}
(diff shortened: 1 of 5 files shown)


soumagne:
Hmm no you don't choose which checksum you use, since you don't create the proc, you just use the one that gets passed to you?

soumagne:
Regarding the checksum footer yes that's true, although to be safe we could consider that it's max 64 bits?

soumagne:
Otherwise looks good to me, you could also add the corresponding HG_Core function that just subtractts the header size from the max unexpected/expected and then have the HG function call the HG Core function and subtract to it the size of the checksum.

soumagne:
XDR uses a fixed size delimited by the eager buf size. XDR encoding is currently broken I think though, I'd need to look at it.

soumagne:
Thanks! Yes that's fine to me, when we add larger checksums we'll reconsider I guess...
One minor thing is that I was thinking that the HG_Core routines would not subtract the checksum size since the user may not use hg_proc, so it would only get subtracted at the HG level...

soumagne:
Yes. Thanks!

soumagne:
Trying the new rebase merge button... let's see :)

soumagne:
The new rebase button seems much better to merge the pull requests, we dropped the pull request number in the commit message this time though but doesn't really matter.

PR #162, opened 2017-03-17
WIP: NA OFI plugin

soumagne:
This is a pull request that we can use to comment / discuss issues etc.


src/na/na_ofi.c
```diff
@@ -0,0 +1,2255 @@
...
+
+    hints->mode               = FI_CONTEXT;
+    hints->ep_attr->type      = FI_EP_RDM;
+    /* notes: verbs provider does not support FI_SOURCE and FI_MR_SCALABLE */
+    hints->caps               = FI_TAGGED | FI_RMA | FI_SOURCE;
+    hints->tx_attr->msg_order = FI_ORDER_SAS;
+    hints->rx_attr->msg_order = FI_ORDER_SAS;
+
+    hints->domain_attr->threading        = FI_THREAD_UNSPEC;
+    hints->domain_attr->control_progress = FI_PROGRESS_MANUAL;
```
sayantansur: Is this required? We don't have an event queue, so the control operations are synchronous. Therefore, the requirement to poll while av_inserts are pending doesn't arise.

soumagne: I would expect in the future though that one event queue gets created per NA context, currently the plugin context_create/context_destroy callbacks are not implemented.

src/na/na_ofi.c
```diff
@@ -0,0 +1,2950 @@
...
+        NA_LOG_ERROR("fi_getinfo failed, rc: %d(%s).", rc, fi_strerror(-rc));
+        fi_freeinfo(hints);
+        ret = NA_PROTOCOL_ERROR;
+        goto out;
+    }
+
+    assert(providers != NULL);
+    nofi_gdata.nog_refcount = 0;
+    hg_thread_mutex_init(&nofi_gdata.nog_mutex);
+    nofi_gdata.nog_providers = providers;
```
jiayulu: na_ofi_getinfo() is called twice, once by na_ofi_check_protoco() i.e na_class_table[plugin_index]->check_protocol, once by na_ofi_initialize() i.e na_private_class->na_class.initialize. the memory returned by the first call is lost.

liuxuezhao: the na_ofi_getinfo() with refcount protection, will only be called once atually.

PR #173, opened 2017-08-09
probe for available BMI port numbers

### LINKED ISSUE #165: HG_Init with BMI without having to specify the port number
When calling HG_Init("bmi+tcp", HG_TRUE), Mercury (actually NA) will always use the default port 22222, which means that 2 servers cannot be started on the same machine without the second failing.

Solution: keep increasing the port number until we find a suitable one.

carns:
Fixes #165 

Requires origin/master version of BMI as of 8/9/2017

- if no port number is specified for BMI in server mode, then try ports
  within a range until one of them works, and store default hostname and
  port as address that can be queried with HG_Addr_to_string() and used
  by other hosts


soumagne:
Merged in af445c7aba71e882f9f18387bea97c7a2577b382

PR #203, opened 2018-01-21
na/ofi: fix a double free problem in error case

hppritcha:
If a provider is unable to initialize properly
when creating an endpoint or domain, the OFI NA
wasn't handling cleanup of OFI info objects quite
right, resulting in heap corruption with double
frees, etc.

Signed-off-by: Howard Pritchard <howardp@lanl.gov>


soumagne:
Merged in 52a20eb

PR #219, opened 2018-04-11
support "verbs;ofi_rxm" libfabric provider

carns:
This allows Mercury to be initialized with "verbs;ofi_rxm" provider combination in addition to the base "verbs" provider.  The former uses a utility provider (ofi_rxm) to implements the RDM endpoint type over the MSG endpoint type of the underlying verbs provider.  This is anticipated to be more stable and performant than the base verbs provider in the long run.

Retains the same flags/options as the existing verbs provider path, but (for now at least) does not mutex lock API calls until we determine if that's needed or not.


soumagne:
Merged in caaaac3

PR #221, opened 2018-04-20
use native psm2 address

jiayulu:
use native psm2 address for the psm2 provider in the ofi plugin.


src/na/na_ofi.c
```diff
@@ -508,28 +508,46 @@ na_ofi_av_insert(na_class_t *na_class, char *node_str, char *service_str,
...
+{
+    /**
+     * http://www.partow.net/programming/hashfunctions/
+     */
+    return (magic * key) ^ ((magic << 3) + (key >> 2));
+}
+
+/* addr_str is in the format: "fi_addr_psmx2://ffff:ff02" */
+static na_uint64_t
+psm2_straddr_2_key(const char *addr_str)
```
liuxuezhao: i have not went through other change, just see this one firstly.
The key has to be able to unique identify an endpoint, any kind of hash cannot fully ensure this.
you may refer some details in PSM2 address encoding to see if can parse/encode to get a unique 64bits key, if cannot, may need to use a 128 bits key.

src/na/na_ofi.c
```diff
@@ -591,6 +604,58 @@ na_ofi_addr_ht_lookup(na_class_t *na_class, struct na_ofi_reqhdr *reqhdr,
...
+                           fi_addr_t *src_addr)
+{
+    na_uint64_t addr_key;
+    fi_addr_t new_src_addr;
+    na_return_t ret;
+
+    addr_key = psm2_straddr_2_key(name);
+
+    /* for psm2 name is in the "psm2://fi_addr_psmx2://40302:0" style */
+    ret = na_ofi_addr_ht_lookup(na_class, name + 7, NULL, addr_key, &new_src_addr);
```
liuxuezhao: I don't grok how it works here? for example the first time you pass "fi_addr_psmx2://40302:0" as node, NULL as service to na_ofi_addr_ht_lookup inside that when not hit hash-table, then na_ofi_av_insert that {node, service}, can the fi_av_insertsvc work for this {node, service}?

liuxuezhao: BTW, did you test with "export FI_PSM2_NAME_SERVER=0" (or unset that ENV), can it works? (better to test with different mode that create multiple SEPs, or one SEP with multiple contexts).

jiayulu: from my experiments fi_av_insertsvc() doesn't work, the fi_getinro()/fi_av_insert() pair works. I tested with export FI_PSM2_NAME_SERVER=0, one SEP with multiple endpoints and multiple basic endpoints work. Did not test multiple SEPs.

PR #235, opened 2018-08-07
bug fixes to fi_writemsg() usage with verbs

### LINKED ISSUE #233: hg_test_read_bw triggers segfault on hg_test_server with verbs;ofi_rxm
The server crashes with a segfault before any performance output is generated.  This is with libfabric 1.6 and mlx5_0 interfaces.  #231 has no impact on behavior.

carns:
- add FI_COMPLETION flag
- fix pointer error in msg_rma.desc (necessary for providers using MR_BASIC)

Fixes #233


soumagne:
Thanks! that looks good to me.

PR #243, opened 2018-09-10
Carns/dev ofi gni fixes

### LINKED ISSUE #241: minor ofi error handling bug
I get the following error:
```
# NA -- Error -- mercury-prefix/src/mercury/src/na/na_ofi.c:1647
 # na_ofi_endpoint_open(): na_ofi_basic_ep_open failed, ret: -28.
```
this "-28" value shouldn't be possible since it's a na_return enum

na_ofi_basic_ep_open() is failing because fi_enable() is returning an error  => there is a mercury error handling bug here around line 1741 that explains the -28 value:
```
    /* Enable the endpoint for communication, and commits the bind operations */
    ret = fi_enable(na_ofi_endpoint->noe_ep);
    if (rc != 0) {
        NA_LOG_ERROR("fi_enable failed, rc: %d(%s).", rc, fi_strerror(-rc));
        ret = NA_PROTOCOL_ERROR;
        goto out;
    }
```
that "ret = fi_enable()" should be "rc = fi_enable(...)" ... this is where the unexpected -28 comes from.

-28 decodes to the OFI version of ENOSPC (No space left on device).

carns:
Fixes #241

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -1736,7 +1736,7 @@ na_ofi_basic_ep_open(const struct na_ofi_domain *na_ofi_domain,
     }
 
     /* Enable the endpoint for communication, and commits the bind operations */
-    ret = fi_enable(na_ofi_endpoint->noe_ep);
+    rc = fi_enable(na_ofi_endpoint->noe_ep);
     if (rc != 0) {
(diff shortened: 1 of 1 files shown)


soumagne:
Thanks Phil, one thing though is that I'm wondering now if we could not remove the use of udreg and just use the internal registration cache. This way we would not even need to show that error message anymore about the NDREG_ENTRIES stuff.

PR #246, opened 2018-09-24
Bulk: bind HG addr to bulk handle and NA OFI addressing fixes

### LINKED ISSUE #207: NA OFI: specify tcp port to use and handle port conflict
If a port is already used it seems that the OFI plugin uses another port that does not correspond to the one that was passed. The port number returned from NA_Addr_to_string is the same though, which is an issue.

soumagne:
HG Bulk: 
- add HG_Bulk_bind(), HG_Bulk_get_addr(), HG_Bulk_get_context_id()
- also add HG_Bulk_bind_transfer() for convenience

HG Bulk proc:
- enable caching of serialized bulk handles / cleanup

HG Core:
- add HG_Core_addr_create() and HG_Core_addr_set_na()
    
NA: 
- add NA_Addr_serialize()/NA_Addr_deserialize()/get_serialize_size() and corresponding plugin callbacks

NA OFI:
- Add na_ofi_addr_serialize() plugin callback implementation
- Rework address handling to only use native addressing format
- Fix passing of init node and domain info in na_ofi_initialize()
- Fix socket provider to bind to specific host/port (fix #207)
- Update gni provider to use FI_SOURCE | FI_SOURCE_ERR caps
- Rework and clean up na_ofi_progress()
- Allocate src_err_addr for fi_cq_readerr()
- Clean up NA OFI op ID and na_ofi_complete()
- Clean up NA OFI plugin and add missing NA_INLINE / prototypes
- Remove unused macros

Testing:
- update bulk test for bind addr testing / clean up return codes

### DIFF
--- a/src/mercury_bulk.c
+++ b/src/mercury_bulk.c
@@ -94,10 +94,13 @@ typedef na_return_t (*na_bulk_op_t)(
 
 /* Note to self, get_serialize_size may be updated accordingly */
 struct hg_bulk {
+    hg_class_t *hg_class;                /* HG class */
     na_class_t *na_class;                /* NA class */
 #ifdef HG_HAS_SM_ROUTING
     na_class_t *na_sm_class;             /* NA SM class */
 #endif
+    hg_uint8_t context_id;               /* Context ID (valid if bound to handle) */
+    hg_core_addr_t addr;                 /* Addr (valid if bound to handle) */
     hg_size_t total_size;                /* Total size of data abstracted */
     hg_uint32_t segment_count;           /* Number of segments */
     struct hg_bulk_segment *segments;    /* Array of segments */
@@ -110,6 +113,8 @@ struct hg_bulk {
     hg_bool_t segment_alloc;             /* Allocated memory to mirror data */
     hg_uint8_t flags;                    /* Permission flags */
     hg_bool_t eager_mode;                /* Eager transfer */
+    void *serialize_ptr;                 /* Cached serialization buffer */
+    hg_size_t serialize_size;            /* Cached serialization size */
     hg_atomic_int32_t ref_count;         /* Reference count */
 };
 
@@ -384,6 +389,7 @@ hg_bulk_create(struct hg_class *hg_class, hg_uint32_t count,
         goto done;
     }
     memset(hg_bulk, 0, sizeof(struct hg_bulk));
+    hg_bulk->hg_class = hg_class;
     hg_bulk->na_class = na_class;
 #ifdef HG_HAS_SM_ROUTING
     hg_bulk->na_sm_class = na_sm_class;
@@ -618,6 +624,10 @@ hg_bulk_free(struct hg_bulk *hg_bulk)
         }
     }
     free(hg_bulk->segments);
+
+    /* Free addr if any was attached to handle */
+    HG_Core_addr_free(hg_bulk->hg_class->core_class, hg_bulk->addr);
+
     free(hg_bulk);
 
 done:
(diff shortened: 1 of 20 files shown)


src/mercury_bulk.c
```diff
@@ -1779,6 +1789,37 @@ HG_Bulk_transfer(hg_context_t *context, hg_cb_t callback, void *arg,
         origin_handle, origin_offset, local_handle, local_offset, size, op_id);
 }
 
+/*---------------------------------------------------------------------------*/
+hg_return_t
+HG_Bulk_bind_transfer(hg_context_t *context, hg_cb_t callback, void *arg,
+    hg_bulk_op_t op, hg_bulk_t origin_handle, hg_size_t origin_offset,
+    hg_bulk_t local_handle, hg_size_t local_offset, hg_size_t size,
+    hg_op_id_t *op_id)
```
liuxuezhao: not sure why need to use another different API? 
it seems that it is easier for user to use if can reuse the same API and differentiate the internal handling by some internal info?

soumagne: Well I would prefer for now that there is a separate API call, it does not seem great otherwise if you have to pass either an addr that is not going to be used or a NULL_ADDR

src/mercury_bulk.c
```diff
@@ -112,12 +114,6 @@ struct hg_bulk {
     hg_uint8_t flags;                    /* Permission flags */
     hg_bool_t eager_mode;                /* Eager transfer */
     hg_atomic_int32_t ref_count;         /* Reference count */
-
-    hg_bool_t bind_addr;                 /* addr binding */
-    void *bulk_hdr;                      /* bulk header */
-    hg_size_t bulk_hdr_size;             /* bulk header size */
```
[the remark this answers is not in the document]
soumagne: Yes sorry I understood what you were doing, I haven't pushed the related change yet (will let you know when I'm done) but instead of doing that I think we can have some better caching done in mercury_proc_bulk.h and simply point to the serialized copy buffer. What I mean is that if the bulk handle was already serialized and you just received it and you intend to forward it again to a different server, we can just do a memcpy of the entire serialized bulk descriptor (not only the bulk header) since the serialized buffer is still available on the receiver's end.

soumagne:
I have reworked the entire NA OFI plugin so that addressing is done using native addressing only, I think it is much cleaner now and we no longer need the fi_getinfo() on lookup. Tests pass for sockets / verbs / psm2. Gni has one issue remaining that seems to be due to FI_SOURCE_ERR not signaling the wait_set correctly, I have to debug that. I also have to enable the "caching" of serialized bulk handles.

soumagne:
We decided that we will also add a HG_Bulk_get_addr() function so that the address can be retrieved from a bulk handle and then passed to HG_Bulk_transfer().

liuxuezhao:
> I have reworked the entire NA OFI plugin so that addressing is done using native addressing only, I think it is much cleaner now and we no longer need the fi_getinfo() on lookup. Tests pass for sockets / verbs / psm2. Gni has one issue remaining that seems to be due to FI_SOURCE_ERR not signaling the wait_set correctly, I have to debug that. I also have to enable the "caching" of serialized bulk handles.

Hi, just curious that after the change, does it support FI_SOURCE? By previous testing, by using FI_SOURCE the PSM2 can get higher RPC rate.

soumagne:
Yes PSM2 still supports FI_SOURCE | FI_SOURCE_ERR, I haven't changed that. The change is that GNI is now also enabled in that mode.

PR #247, opened 2018-09-28
fix uninitialized variable

carns:


soumagne:
Ah I see thanks I'll just do a memset instead to make sure the struct is initialized

carns:
Ok, sounds good.  I found this while trying to track down a problem in a different WIP merge request; once you have a memset fix in master I'll rebase that branch.

soumagne:
It's in master now

carns:
thanks Jerome!

PR #260, opened 2018-11-05
enable wait_fd and auto progress for psm2

carns:
Tested on Bebop system at ANL/LCRC.  This sets the WAIT_FD option in na_ofi in combination with the AUTO progress mode (the former will not work without the latter).

Causes a penalty to latency but not throughput in exchange for reducing CPU consumption when not in busy spin mode.


soumagne:
Ah I just see your pull request :( after I checked it in already... OK well at least on the positive side, we reached the same conclusion :)

soumagne:
Merged in 865068c

PR #275, opened 2019-02-20
return error in na_ofi_cq_read

jiayulu:
let na_ofi_cq_read return NA_PROTOCOL_ERROR instead of NA_SUCCESS if the
error code is FI_EIO.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -2764,6 +2764,7 @@ na_ofi_cq_read(na_class_t *na_class, na_context_t *context,
                 cq_err.err, fi_strerror(cq_err.err),
                 cq_err.prov_errno,
                 fi_strerror(-cq_err.prov_errno));
+            ret = NA_PROTOCOL_ERROR;
             goto out;
         default:
             NA_LOG_ERROR("fi_cq_readerr got err: %d(%s), "


soumagne:
thanks! I agree it should return that. I can't remember now though if there was any specific reason for not returning NA_PROTOCOL_ERROR, have you checked with @liuxuezhao ?

src/na/na_ofi.c
```diff
@@ -2764,6 +2764,7 @@ na_ofi_cq_read(na_class_t *na_class, na_context_t *context,
                 cq_err.err, fi_strerror(cq_err.err),
                 cq_err.prov_errno,
                 fi_strerror(-cq_err.prov_errno));
+            ret = NA_PROTOCOL_ERROR;
```
liuxuezhao: not sure why you need to change it here?

The reason of why did not set is as error before is it happens in the case that when killing a target the the in-progress communication and cq read probably get this error, but if return it as failure then the NA_Progress will get an error and make the caller stop progress. So it just print a msg and ignore the EIO so user can continue to make progress.  
This change will cause some problems when killing some targets while communication in progress.

yulujia: we saw this: process B exited, then process A sends RPC to process B. HG_Forward() returned HG_SUCCESS, but it actually failed. We are relying on the return code from HG_Forward() for error handling.

liuxuezhao: if B exited ahead, and then A send RPC to B, in this case the fi_tsend() should return error.
in which test case did you meet that?

If B's exiting/dead happens in race condition with A send RPC to B, in DAOS test cases we ever got the EIO in fi_cq_read() so the original code ignored that error code so it will not cause NA_Progress' error. And the RPC will finally get timeout.

liuxuezhao:
BTW, for your error handling (in cart or IOF), simply relying on the return code from HG_Forward() for error handling probably not enough, in some case the fi_tsend()'s failure cannot be catched, better to check if its completion callback is called.

src/na/na_ofi.c
yulujia: we saw that in the test_corpc_version test with 32 processes. We did see the RPC time out later on.

liuxuezhao: The RPC timeout should be able to be used to check its failure.
Even with this patch, you cannot get the HG_Forward()'s error as the fi_tsend succeed, but the NA_Progress will fail.

soumagne:
Fixed in b4e7cb221ec3cdde1f233a3e9febe16263552136

PR #284, opened 2019-04-09
fix na_ofi_cancel

yulujia:
call na_ofi_complete() even when fi_cancel() fails.


soumagne:
@yulujia could you please remind me which libfabric provider you're testing against? is it only sockets or you also see the issue with psm2?

yulujia:
@soumagne , I only see the problem with the sockets provider, psm2 seems fine.

yulujia:
in the original code if fi_cancel() fails, ret is still returned as NA_SUCCESS, caller of HG_Cancel can't detect the failure.

PR #286, opened 2019-05-15
fix bug in header encoding for null input

carns:
- Don't skip too many header encoding steps if input struct and/or encoder is null.
- This fix ensures that Mercury includes the HG_Class_set_input_offset() area when transmitting RPCs with no input arguments.

### DIFF
--- a/src/mercury.c
+++ b/src/mercury.c
@@ -657,7 +657,7 @@ hg_set_struct(struct hg_private_handle *hg_handle,
     }
     if (!proc_cb || !struct_ptr) {
         /* Silently skip */
-        *payload_size = 0;
+        *payload_size = header_offset;
         goto done;
     }
 


src/mercury.c
```diff
@@ -655,11 +655,6 @@ hg_set_struct(struct hg_private_handle *hg_handle,
             ret = HG_INVALID_PARAM;
             goto done;
     }
-    if (!proc_cb || !struct_ptr) {
-        /* Silently skip */
-        *payload_size = 0;
```
soumagne: Instead of the changes that you've made, could you please try to change that line with `*payload_size = header_offset`

soumagne:
Also I assume we need the same fix for the output.

src/mercury.c
carns: Sure thing.  I tested that suggestion and it works fine; PR has been updated accordingly.

soumagne: Great thanks!

PR #290, opened 2019-06-05
fix refcounting in na_ofi_addr_lookup

yulujia:
increment na_ofi_op_id->noo_refcount by 1 instead of setting it to 1.


soumagne:
@yulujia this creates memory leaks, see https://cdash.hdfgroup.org/viewDynamicAnalysis.php?buildid=27404 I think instead of incrementing the refcount we should make sure that the refcount is 1 at this point.

yulujia:
@soumagne if the code enters teh branch here:
https://github.com/mercury-hpc/mercury/pull/290/files#diff-3c6058bbbf011335ba35366df9515df7L3649

won't line 3668 lose refcount?

soumagne:
ah right, I think that line should just be deleted then. We actually currently never go into that situation because we don't re-use op IDs for address lookups.

yulujia:
fixed by f0b9f992793be46f1c6ae47b30d1c3ccb525cfbf

PR #300, opened 2019-08-16
fix premature assignment of na_op_id

yulujia:
Do not assign pointer if the newly allocated op_id struct is freed.
Otherwise the pointer will point to garbage values.


yulujia:
@soumagne @carns what do you think of this?

carns:
This looks like a good change to me.

soumagne:
@yulujia thanks I will take a closer look. I think we could even get rid of these assignments since most of them are never used because the OP ID is pre-allocated for all operations except for the lookup.

soumagne:
Fixed in b4e7cb221ec3cdde1f233a3e9febe16263552136

PR #303, opened 2019-08-23
input offset bugfix

### LINKED ISSUE #302: extra input buffer fields not transmitted correctly for large RPC requests
This problem was discovered and isolated by @srini009 (thanks!), I'm just contributing a test case to demonstrate it in the Mercury source tree.

**Describe the bug**

Mercury allows callers to reserve and access additional space in RPC requests using the HG_Class_set_input_offset() and HG_Get_input_buf() functions.  This works correctly for normal-sized RPC input structs, but if the input is large enough to trigger rendezvous mode (in which the encoded input struct transfered with a bulk (rdma) transfer), then the extra information is not transmitted correctly.

**To Reproduce**

I apologize that I don't really understand how to create new ctest test cases quickly, but here is a modification to the first existing rpc test case that makes it just loop through rpcs, increasing the size of a string transmitted in the request, 100 bytes at a time, until an incorrect value is detected in the HG_Get_input_buf() area.

https://github.com/carns/mercury/commit/ac3dffab78514218be558746e74e3bb8455ad54b

https://github.com/carns/mercury/tree/carns/dev-input-offset-test

The server will produce this output with na+sm:
```Got tester value: 123456
Got tester value: 123456
Got tester value: 123456
Got tester value: 123456
Got tester value: 123456
Got tester value: 123456
Got tester value: 4100
Incorrect value in extra header field.
```
While the client produces this output:
```
Sending RPC that includes streng of length 3600 in input struct.
 PASSED
Sending RPC that includes streng of length 3700 in input struct.
 PASSED
Sending RPC that includes streng of length 3800 in input struct.
 PASSED
Sending RPC that includes streng of length 3900 in input struct.
 PASSED
Sending RPC that includes streng of length 4000 in input struct.
 PASSED
Sending RPC that includes streng of length 4100 in input struct.
```
Everything hangs at that point, probably because I don't know how to correctly report an error on the server side in the test cases.

**Expected behavior**

Test case should complete successfully regardless of size of input being transmitted.

**Platform (please complete the following information):**

Linux, using mercury origin/master, with na+sm transport.

**Additional context**

carns:
Fixes #302 

The encoding of extra input header information was correct, it's just that the HG_Get_input_buf() function assumed that it had been moved to the extra_buf memory buffer if the request exceeded the eager limit.

This PR makes the encoder (hg_set_struct()) and the function for accessing the extra header (HG_Get_input_buf()) consistent in assuming the data is in the request.

It also adds a safety check to the HG_Class_set_input_offset(), because the extra header information cannot be larger than the eager buffer size.

### DIFF
--- a/src/mercury.c
+++ b/src/mercury.c
@@ -1971,6 +1971,8 @@ HG_Get_input_buf(hg_handle_t handle, void **in_buf, hg_size_t *in_buf_size)
     struct hg_private_handle *private_handle =
         (struct hg_private_handle *) handle;
     hg_return_t ret = HG_SUCCESS;
+    void *buf;
+    hg_size_t buf_size, header_offset = hg_header_get_size(HG_INPUT);
 
     if (!handle) {
         HG_LOG_ERROR("NULL HG handle");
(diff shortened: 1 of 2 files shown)


soumagne:
Thanks!, I think I see the problem now, I think maybe we should have added 2 routines instead of trying to handle everything in Get_input_buf(), i.e. we should have get_input_buf and get_extra_input_buf()

soumagne:
I will clean it up

soumagne:
actually yes I think you're right, it's probably enough to limit to the eager message size, I'll just do the same for the output.

soumagne:
Merged in c35f69029e75c291ccabb6349b501cb04aab8c31

PR #311, opened 2019-09-26
use fallback when getlogin() fails

carns:
If getlogin() fails on a system, use "unknown" as a placeholder string
to construct the /tmp paths to fall back gracefully.


soumagne:
I merged that fix in b9bf0654c858f845b237a67a40402804db8f3417 I actually changed the `getlogin` call to instead use `getpwuid` which seems safer.

carns:
great, thanks!

PR #316, opened 2019-10-05
NA RECV: force hg_core_handle->ret to HG_CANCELED

yulujia:
for na recv operatoins, always set hg_core_handle->ret to HG_CANCELED
when HG_Cancel() didn't return error.


soumagne:
No longer needed after changes in #330

PR #318, opened 2019-10-24
NA OFI: use ip addr for psm2

yulujia:
let psm2 uses ip:port address format so that user can specify uri at
init time.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -346,6 +346,8 @@ struct na_ofi_context {
 /* Endpoint */
 struct na_ofi_endpoint {
     struct na_ofi_addr *src_addr;           /* Endpoint address         */
+    char *src_node;                         /* Fabric address           */
+    char *src_service;                      /* Service name             */
     struct fi_info *fi_prov;                /* Provider info            */
     struct fid_ep *fi_ep;                   /* Endpoint handle          */
     struct fid_wait *fi_wait;               /* Wait set handle          */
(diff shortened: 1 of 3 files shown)


yulujia:
@soumagne could you comment on this PR? I wasn't sure what you meant in the email.

src/na/na_ofi.c
```diff
@@ -2600,13 +2703,37 @@ na_ofi_cq_read(na_context_t *context, size_t max_count,
...
+            void *err_addr = NULL;
+            size_t err_addrlen;
+
+            /* Copy addr information */
+            err_addr = malloc(cq_err.err_data_size);
+            NA_CHECK_ERROR(err_addr == NULL, out, ret, NA_NOMEM_ERROR,
+                "Could not allocate err_addr");
+            err_addrlen = cq_err.err_data_size;
+            memcpy(err_addr, cq_err.err_data, err_addrlen);
+
```
soumagne: I don't think that block of code is needed anymore, I removed it previously so that we don't need the extra malloc. The av insert is done in the unexpected event callback.

yulujia: without this, the na_ofi_cq_process_recv_unexpected_event() function will enter the branch: ```else if (src_err_addr && src_err_addrlen)```, and in there ```src_err_addr``` contains the psm2 internal service id,  then the server will fail on line 1287, because that fi_getinfo() needs the user-passed-in fake service id, but src_err_addr contains the psm2 internal service id. I didn't find a way to get around that.

soumagne: Hmm ok I'll take a look. It should be fixed. Also you're inserting directly into the AV which is probably not right, it should go through the hash table stuff.

yulujia: > Also you're inserting directly into the AV which is probably not right
You are right. the address should be inserted to the hash table.  On the client side I packed the ip and port on line 2280, is there a way to retrieve that on the server side for the recv_unexpected_event case? or do we need to use a req header like the sockets provider?

soumagne: I don't think you should use the req header. Do you really need to pack the IP/port to get that to work?

yulujia: that's the only way I can think of for the client to pass its ip:port uri to the server. The server passes its URI to the client by the port.cfg file, the client can't do it that way. Do you have any suggestions?

soumagne:
closing as we now handle multiple address formats for each provider

Issue #308, opened 2019-09-09
NA OFI: strange HG_Cancel() behaviour


yulujia:
using mercury commit >= b65ccb3066435 I experience some strange behavior with HG_Cancel().

case 1:
1) start to service procs proc_a and proc_b, each creates a NA context, and a dedicated progress thread.
2) proc_a sends RPC to proc_b, sleeps for 5 seconds, call HG_Cancel() to cancel the RPC. HG_Cancel() fails with HG_CANCEL_ERROR
3) proc_a fails during HG_Context_destroy(), error message:

09/03-23:47:22.31 CaRT[26606/26606] hg ERR # NA -- Error -- mercury/src/na/na_ofi.c:4790
na_ofi_cancel(): fi_cancel unexpected recv failed, rc: -2(No such file or directory).
09/03-23:47:22.31 CaRT[26606/26606] hg ERR # HG -- Error -- mercury/src/mercury_core.c:991
hg_core_pending_list_cancel(): Could not cancel handle
09/03-23:47:22.31 CaRT[26606/26606] hg ERR # HG -- Error -- mercury/src/mercury_core.c:3836
HG_Core_context_destroy(): Cannot cancel list of pending entries
09/03-23:47:22.31 CaRT[26606/26606] hg ERR # HG -- Error -- mercury/src/mercury.c:1230
HG_Context_destroy(): Could not destroy HG core context

case 2:
1) start to service procs proc_a and proc_b, each creates a NA context, and a dedicated progress thread.
2) proc_a sends RPC to proc_b, sleeps for 5 seconds, call HG_Cancel() to cancel the RPC. HG_Cancel() fails with HG_CANCEL_ERROR
3) call HG_Destroy() multiple times until the RPC handle is destroyed.
4) Later on, the RPC completion callback is triggered anyway, and we get a segfault since the hg_handle is already destroyed.

dmiter:
Mercury code assumes that the `fi_cancel` operation is blocking and will always actually cancel the operation. These assumptions are incorrect.

Here is the cancel text from the libfabric man page:

>fi_cancel
[...]

PR #327, opened 2019-11-26
NA OFI: ignore fi_cancel's failure

liuxuezhao:
Ignore fi_cancel's failure (just add a warning log msg).
We observed a DAOS test failure that:
1) RPC timedout for some reason,
2) DAOS calls HG_Cancel() but it fails, then DAOS complete that RPC
3) The RPC's completion callback (crt_hg_req_send_cb) being triggered
   again by mercury, this cause the error that one RPC's completion
   callback being called twice.

This patch fixes it by ignoring fi_cancel()'s failure, so won't return
error by HG_Cancel() by step 2), because at mercury level it return
error for HG_Cancel() and trigger the same RPC's completion callback
later should be an issue.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -4413,9 +4413,9 @@ na_ofi_cancel(na_class_t *na_class, na_context_t *context,
 
         rc = fi_cancel(&NA_OFI_CONTEXT(context)->fi_rx->fid,
             &na_ofi_op_id->fi_ctx);
-        NA_CHECK_ERROR(rc != 0, out, ret, NA_CANCEL_ERROR,
-            "fi_cancel() unexpected recv failed, rc: %d(%s)",
-            rc, fi_strerror((int) -rc));
+        if (rc != 0)
+            NA_LOG_WARNING("fi_cancel() unexpected recv failed, rc: %d(%s)",
+                           rc, fi_strerror((int) -rc));
 
         tmp = first = na_ofi_msg_unexpected_op_pop(context);
         do {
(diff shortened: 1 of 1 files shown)


liuxuezhao:
We met the problem in DAOS rebuild test, and have a cart ticket https://jira.hpdd.intel.com/browse/CART-820.
@soumagne could you please check if this patch is reasonable to you? thanks

carns:
FWIW this seems sensible to me. Mercury informing the transport layer that it should cancel seems like it should be a "best effort" kind of thing, especially since we can't know the true outcome of the cancellation attempt anyway until we get notification from libfabric for completion of the operation in question.

soumagne:
Yes that also seems fine with me. Probably seems fine also wrt previous #308 comments.

soumagne:
I will take a closer look and will make sure we can get that in and close all the cancel related PRs that we have pending.

liuxuezhao:
Thanks you both, just FYI that WangDi verified that DAOS rebuild test can pass  with this patch.

liuxuezhao:
> I will take a closer look and will make sure we can get that in and close all the cancel related PRs that we have pending.

Hi Jerome, any conclusion for this issue?
What we observed in DAOS/CART (log available at https://jira.hpdd.intel.com/browse/CART-820), we firstly get failure from HG_Cancel() with return code of HG_CANCEL_ERROR, that means the cancel failed so DAOS has to explicitly complete the RPC with an error code. But right after that, that RPC's completion callback being triggered (by calling HG_progress/trigger) with hg_cb_info::ret set as HG_CANCELED and then complete the RPC another time.
With this patch can fix the issue.

soumagne:
@liuxuezhao yes sorry I'm almost done with the previous fix. I'm planning to merge that one just after.

liuxuezhao:
> @liuxuezhao yes sorry I'm almost done with the previous fix. I'm planning to merge that one just after.

No problem, I am fine that if other alternative solution can make it works. Is there a PR for that? or maybe have not pushed yet.

soumagne:
No longer needed after changes in #330

PR #347, opened 2020-02-27
ofi: complete rpc if fi_cancel() fails

yulujia:
if fi_cancel() returns error, call na_ofi_complete() to ensure rpc
completion.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -4468,13 +4468,14 @@ na_ofi_cancel(na_class_t *na_class, na_context_t *context,
      * or it will show up in the regular completion queue.
      */
     rc = fi_cancel(&fi_ep->fid, &na_ofi_op_id->fi_ctx);
-    NA_LOG_DEBUG("fi_cancel() rc: %d(%s)", (int) rc,
-        fi_strerror((int) -rc));
+    NA_LOG_DEBUG("fi_cancel() rc: %d(%s)", (int) rc, fi_strerror((int) -rc));
 //    NA_CHECK_ERROR(rc == -FI_ENOENT, out, ret, NA_OPNOTSUPPORTED,
 //        "fi_cancel() failed, rc: %d(%s)", rc, fi_strerror((int) -rc));
-    if (rc != 0) {
+    if (rc == FI_ENOENT) {
         NA_LOG_DEBUG("Operation ID %p failed to cancel", na_ofi_op_id);
-        ret = na_ofi_complete(na_ofi_op_id, NA_CANCELED);
+        if (hg_atomic_cas32(&na_ofi_op_id->status, NA_OFI_OP_CANCELED,
+                            NA_OFI_OP_COMPLETED))
+            ret = na_ofi_complete(na_ofi_op_id, NA_CANCELED);
     }
 
     /* Work around segfault on fi_cq_signal() in some providers */
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -4434,6 +4434,7 @@ na_ofi_cancel(na_class_t *na_class, na_context_t *context,
     struct na_ofi_op_id *na_ofi_op_id = (struct na_ofi_op_id *) op_id;
     struct fid_ep *fi_ep = NULL;
     na_return_t ret = NA_SUCCESS;
+    na_bool_t manual_complete = NA_FALSE;
     ssize_t rc;
 
     /* Exit if op has already completed */
@@ -4451,6 +4452,7 @@ na_ofi_cancel(na_class_t *na_class, na_context_t *context,
             break;
         case NA_CB_SEND_UNEXPECTED:
         case NA_CB_SEND_EXPECTED:
+            manual_complete = NA_TRUE;
         case NA_CB_PUT:
         case NA_CB_GET:
             fi_ep = NA_OFI_CONTEXT(context)->fi_tx;
@@ -4471,11 +4473,12 @@ na_ofi_cancel(na_class_t *na_class, na_context_t *context,
     NA_LOG_DEBUG("fi_cancel() rc: %d(%s)", (int) rc, fi_strerror((int) -rc));
 //    NA_CHECK_ERROR(rc == -FI_ENOENT, out, ret, NA_OPNOTSUPPORTED,
 //        "fi_cancel() failed, rc: %d(%s)", rc, fi_strerror((int) -rc));
-    if (rc == FI_ENOENT) {
-        NA_LOG_DEBUG("Operation ID %p failed to cancel", na_ofi_op_id);
-        if (hg_atomic_cas32(&na_ofi_op_id->status, NA_OFI_OP_CANCELED,
-                            NA_OFI_OP_COMPLETED))
-            ret = na_ofi_complete(na_ofi_op_id, NA_CANCELED);
+    /* in sockets provider if fi_ep is a TX context fi_cancel() returns
+     * -FI_NOENT right away.
+     */
+    if (rc == -FI_ENOENT && manual_complete) {
+        NA_LOG_WARNING("Operation ID %p failed to cancel", na_ofi_op_id);
+        ret = na_ofi_complete(na_ofi_op_id, NA_CANCELED);
     }
 
     /* Work around segfault on fi_cq_signal() in some providers */
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -4437,8 +4437,10 @@ na_ofi_cancel(na_class_t *na_class, na_context_t *context,
     ssize_t rc;
 
     /* Exit if op has already completed */
-    if (!hg_atomic_cas32(&na_ofi_op_id->status, 0, NA_OFI_OP_CANCELED))
+    if (!hg_atomic_cas32(&na_ofi_op_id->status, 0, NA_OFI_OP_CANCELED)) {
+        NA_LOG_DEBUG("Operation ID %p already completed", na_ofi_op_id);
         goto out;
+    }
 
     NA_LOG_DEBUG("Canceling operation ID %p", na_ofi_op_id);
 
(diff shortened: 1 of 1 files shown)


soumagne:
@yulujia  What type of error do you get from fi_cancel()?

yulujia:
> @yulujia What type of error do you get from fi_cancel()?

we saw that fi_cancel() failed on this line:
```https://github.com/mercury-hpc/mercury/pull/347/files#diff-3c6058bbbf011335ba35366df9515df7R4470```

then HG_Cancel() returns success, mercury never executes the rpc completion callback.

**Comment by soumagne** [comment]
@yulujia which error code from fi_cancel() did you get? we need to make sure that this is an expected error code from libfabric and not simply create a patch to fix a test. Can you please also detail your use case. This fix is not correct because the operation may have already completed so you are creating a race by adding immediate completion. Either we need to test against a specific fi_cancel() error code or we need to fix libfabric but we cannot complete the OP immediately like that.

**Comment by yulujia** [comment]
@soumagne the error code was -2 (No such file or directory). My test is doing this:
1) launch two processes, rank_0 and rank_1
2) rank_1 tells rank_0 to exit
3) rank_1 sends rank_0 an RPC, wait for the RPC to timeout, then cancel the RPC. This is when we hit the error. 

The man page for fi_cancel says:
```
 For fi_cancel, a return value of 0 indicates that the cancel request was submitted for processing.
```
that implies an error code means the cancel request wasn't submitted. In that case it shouldn't introduce a race condition.

logs messages around the error:
```
5092 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] rpc  ERR  src/cart/crt_context.c:763 crt_req_timeout_hdlr(0x7f70f800b010)  [opc=0xfe000000 xid=0x1a rank:tag=0:0] aborting to group service_group, rank 0, tgt_ur  (null)
   1 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] hg   DBUG # NA -- Debug -- /home/yulujia/codes/cart-hg-cancel/_build.external-Linux/mercury/src/na/na_ofi.c:4392                                                  
   2  # na_ofi_cancel(): Operation ID 0x7f70f800b380 already completed                                                                                                                                             
   3 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] hg   DBUG # NA -- Debug -- /home/yulujia/codes/cart-hg-cancel/_build.external-Linux/mercury/src/na/na_ofi.c:4396                                                  
   4  # na_ofi_cancel(): Canceling operation ID 0x7f70f800af10                                                                                                                                                     
   5 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] hg   DBUG # NA -- Debug -- /home/yulujia/codes/cart-hg-cancel/_build.external-Linux/mercury/src/na/na_ofi.c:4423                                                  
   6  # na_ofi_cancel(): fi_cancel() rc: -2(No such file or directory)                                                                                                                                             
   7 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] hg   ERR  # NA -- Error -- /home/yulujia/codes/cart-hg-cancel/_build.external-Linux/mercury/src/na/na_ofi.c:4427                                                  
   8  # na_ofi_cancel(): Operation ID 0x7f70f800af10 failed to cancel                                                                                                                                              
   9 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] hg   DBUG # NA -- Debug -- /home/yulujia/codes/cart-hg-cancel/_build.external-Linux/mercury/src/na/na_ofi.c:4392                                                  
  10  # na_ofi_cancel(): Operation ID 0x7f70f800b430 already completed                                                                                                                                             
  11 02/27-10:16:31.02 wolf-21 CaRT[56739/56800] rpc  DBUG src/cart/crt_context.c:816 crt_context_timeout_check(0x7f70f800b010)  [opc=0xfe000000 xid=0x1a rank:tag=0:0] decref to 2.                               ```

carns:
Re: potential race conditions, what if the operation were to successfully complete after a failed cancellation attempt?  That seems like the danger here.  I don't think the caller here would be able to guarantee that won't happen, and if it did, then the network operation would access memory that the Mercury API (by way of the callback) had already indicated was safe for reuse.

yulujia:
@carns I see what you mean. In the case I described above what would be a proper way to ensure the rpc completion callback is triggered?

carns:
I'm not sure what Mercury could do to help in this scenario.  Any idea why fi_cancel() failed?  That seems like the underlying problem that really needs to be fixed to handle this safely.

soumagne:
@yulujia right agree with @carns that was exactly my point. If the operation already succeeded you will get a return code of -FI_ENOENT, that's why I had that debugging code there before. We cannot rely on that return code to take any action but only on the events that get generated from libfabric. There are two resulting code paths currently:
- https://github.com/mercury-hpc/mercury/blob/master/src/na/na_ofi.c#L2564 when fi_cancel() succeeds and we get an FI_ECANCELED event
- https://github.com/mercury-hpc/mercury/blob/master/src/na/na_ofi.c#L2791 where cancelation may have occurred while the request was being processed and _before_ it was completed, in which case fi_cancel() would return FI_ENOENT

Either way those two code paths should currently be the only ones that can be taken. If we don't fall in either of those code paths then we should understand why first, before making any changes to the code or trying to complete a request immediately (which can be unsafe). Currently I do not see a reason for your use case to not fall in those two code paths. Are you making progress after you call HG_Cancel()?

yulujia:
> Are you making progress after you call HG_Cancel()?

Yes the progress thread is still running.

> Either way those two code paths should currently be the only ones that can be taken.

In both code paths you assume fi_cancel() successfully submitted the cancel request. In the error case I had, it looks to me fi_cancel() didn't submit the request in the first place, so neither of the two code paths were taken.

soumagne:
No, I just said that in the second bullet you would get a FI_ENOENT return code. In any case, if you get a FI_ENOENT return code, you need to understand why, that means that the operation cannot be found, so where is it? you cannot simply immediately complete the operation without knowing where the operation is gone.

yulujia:
> n any case, if you get a FI_ENOENT return code ... that means that the operation cannot be found, so where is it?

the only thing regarding return code in the man page is this:
``` 
For fi_cancel, a return value of 0 indicates that the cancel request was submitted for processing.
```
on the flip side, a non zero return code means the request was not submitted. There is no explicit explanation about different error codes, how do you know FI_ENOENT means the operation cannot be found? If the operation is not processed by fi_cancel(), it should be still sitting in the same place before fi_cancel() is called .

soumagne:
You should have a look at the libfabric code, the man page is not sufficient. I can just guarantee you that your current fix is prone to races. If the operation is successfully submitted you cannot assume that the operation is sitting in the same place. If it were, then fi_cancel() would return 0. Now it's also possible that there is a problem in libfabric and we need to fix it but we should make sure that we understand it or understand what is happening, you cannot immediately complete an operation without knowing what is happening or by simply making assumptions.

Also does this happen with all the providers or is it only with one specific provider?

yulujia:
Okay I took a look at the libfabric code. For the sockets provider fi_cancel() returns -FI_ENOENT right away if the fid passed in is a send:
```
534          case FI_CLASS_TX_CTX:                                                                                                                                                                                 
   1         case FI_CLASS_STX_CTX:                                                                                                                                                                                                                                                                                       
   2                 return -FI_ENOENT;
```
this patch will not create a race condition. If the code reaches the point at the new na_ofi_complete() call, it not possible that the op_id can complete anywhere else.

I am not sure about other providers. The cancel code for rxm seems to do the right thing.

soumagne:
no I'll try to explain again...
1. you are trying to cancel a send and you don't know if the target is alive or not.
2. as soon as the send is posted, there is a chance that it can successfully complete
3. you cannot complete the operation ID there because there is a chance that this OP ID will show up in the completion queue. What you are doing is equivalent to calling free() on something that is still in use.

So either we need to fix libfabric or maybe there is another error code in the completion queue that we need to catch (which I can look at) but your fix as it is will create issues. Please send me your test code so that I can make sure that it's doing the right thing.

yulujia:
> you cannot complete the operation ID there because there is a chance that this OP ID will show up in the completion queue. What you are doing is equivalent to calling free() on something that is still in use.

If you check for NULL pointer before calling free() it won't be problem. When na_ofi_progress() retrieves the FI_ECANCELED event, it checks for NA_OFI_OP_CANCELED, that's equivalent to checking for NULL pointer.

Given the status of fi_cancel(), forcing completion in na_ofi_cancel() works better than the way it currently is. See https://github.com/mercury-hpc/mercury/pull/327

Also, HG_Cancel() doesn't work as expected on verbs. If rank 1 sends an RPC to an already exited rank 0, the rpc completion callback is never executed. 

my test is i this PR: https://github.com/daos-stack/cart/pull/361
the test is named: test_rpc_to_ghost_rank
after cart is built, the way we run it is:
edit
```
    install/Linux/TESTING/corpc/cart_corpc_one_node.yaml 
```
so that it only contains the ghost_rank test. Also replace boro-A with your hostname. then:
```
    cd /install/Linux/TESTING
    ./launch.py -s one_node,corpc
```

Another test that manifests the HG_Cancel() issue is:
```
    ./launch.py -s one_node,iv
```

carns:
@yulujia , the scenario you described in the above comment is not thread safe.  Consider a configuration in which a dedicated thread has been assigned to running the Mercury progress loop, and another set of threads is servicing RPCs.  The fi_cancel() and na_ofi_progress() calls will always be issued by different threads in that configuration.

yulujia:
@carns, there's the atomic compare and set operation, that will make it thread-safe, right?

soumagne:
@yulujia #327 is wrong too and was an attempt to fix the original code that was not correct. You say:
> Given the status of fi_cancel(), forcing completion in na_ofi_cancel() works better than the way it currently is.

Sorry but that's not the way we should work, we need to identify and fix the problems at the source and in libfabric if there's a problem there but we certainly do not introduce workarounds that create races. You need to understand that as soon as you call `na_ofi_complete()`, that pushes the operation ID into the completion queue and makes it ready to be re-used. If later on, libfabric tells us that the operation completed after an `fi_cq_read()`, what will happen is that the pointer to the operation ID that will be retrieved will be invalid/freed/re-used and we'll have no safe way to check. So again, it's not only about the atomic operation, the atomic operation only protects the accesses to the "status" field of the operation ID. If you release the operation ID and then later on, the pointer to that operation ID shows up (because the send() may still successfully complete, we just can't know), it will be an invalid read.

soumagne:
Closing that PR as this is a libfabric issue now tracked in https://github.com/ofiwg/libfabric/issues/5701

PR #350, opened 2020-03-04
Add initial MPI_T-like profiling support in Mercury

srini009:
Add MPI-like profiling interface in Mercury. 
Profiling interface is active by default. 
Currently, only one PVAR is exported called "hg_pvar_hg_forward_count" that counts the number of times the HG_Forward call is invoked on this instance.

### DIFF
--- a/src/mercury.c
+++ b/src/mercury.c
@@ -22,6 +22,10 @@
 #include <string.h>
 #include <assert.h>
 
+/* PVAR profiling support */
+#include "mercury_prof_pvar_impl.h"
+HG_PROF_PVAR_UINT_COUNTER_DECL_EXTERN(hg_pvar_hg_forward_count);
+
 /****************/
 /* Local Macros */
 /****************/
@@ -959,6 +963,7 @@ HG_Init_opt(const char *na_info_string, hg_bool_t na_listen,
     const struct hg_init_info *hg_init_info)
 {
     struct hg_private_class *hg_class = NULL;
+    int ret;
 #ifdef HG_HAS_VERBOSE_ERROR
     const char *log_level = NULL;
 
@@ -986,6 +991,11 @@ HG_Init_opt(const char *na_info_string, hg_bool_t na_listen,
     /* Set more data callback */
     HG_Core_set_more_data_callback(hg_class->hg_class.core_class,
         hg_more_data_cb, hg_more_data_free_cb);
+
+
+    /* EXPERIMENTAL: Initialize PVAR profiling data structures */
+    ret = hg_prof_pvar_init();
+    assert(ret == HG_SUCCESS);
 
     return (hg_class_t *) hg_class;
 
@@ -1917,6 +1927,10 @@ HG_Forward(hg_handle_t handle, hg_cb_t callback, void *arg, void *in_struct)
         goto done;
     HG_CHECK_HG_ERROR(done, ret, "Could not forward call (%s)",
         HG_Error_to_string(ret));
+
+    /* PVAR profiling support: Increment the value of the PVAR with the name "hg_pvar_hg_forward_count" */
+    fprintf(stderr, "[MERCURY_PROF_INTERFACE] Incrementing the value of PVAR with name hg_pvar_hg_forward_count\n");
+    HG_PROF_PVAR_COUNTER_INC(hg_pvar_hg_forward_count, 1);
 
 done:
     return ret;
--- a/src/mercury_prof_interface.c
+++ b/src/mercury_prof_interface.c
@@ -1,0 +1,198 @@
+/*
+ * Copyright (C) 2013-2020 Argonne National Laboratory, Department of Energy,
+ *                    UChicago Argonne, LLC and The HDF Group.
+ * All rights reserved.
+ *
+ * The full copyright notice, including terms governing use, modification,
+ * and redistribution, is contained in the COPYING file that can be
+ * found at the root of the source code distribution tree.
+ */
+
+#include "mercury_bulk.h"
+#include "mercury_core.h"
+#include "mercury_private.h"
+#include "mercury_error.h"
+
+#include "mercury_atomic.h"
+#include "mercury_prof_interface.h"
+#include "mercury_prof_pvar_impl.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
+
+/* Variable denoting whether or not the profiling interface has been initialized */
+static int hg_prof_is_initialized = 0;
+
+/* Client-side PVAR handle that contains all information about the PVAR. Data structure is opaque to the client. 
+ * Handle exists to make PVAR reads quicker */
+struct hg_prof_pvar_handle {
+   hg_prof_class_t pvar_class;
+   hg_prof_datatype_t pvar_datatype;
+   hg_prof_bind_t pvar_bind;
+   int continuous;
+   void * addr;
+   int is_started;
+   int count;
+   char name[128];
+   char description[128];
+};
+
+/* Client-side PVAR session. PVARs are associated with a session. As of now, we only support one default session */
+struct hg_prof_pvar_session {
+   hg_prof_pvar_handle_t * pvar_handle_array;
+   int num_pvars;
+   int reference_counter;
+};
+
+struct hg_prof_pvar_session default_session;
+
+/* Setter and getter functions for profiling interface initialization */
+static void hg_prof_set_is_initialized(int val)
+{
+  hg_prof_is_initialized = val;
+}
+
+static int hg_prof_get_is_initialized() {
+   return hg_prof_is_initialized;
+}
+
+/* Client-side API to initialize the PVAR profiling interface */
+hg_return_t HG_Prof_init() {
+
+  default_session.reference_counter = 0;
+  default_session.num_pvars = NUM_PVARS;
+  default_session.pvar_handle_array = (hg_prof_pvar_handle_t *)malloc(sizeof(hg_prof_pvar_handle_t)*NUM_PVARS);
+  hg_prof_set_is_initialized(1);
+
+  return HG_SUCCESS;
+}
+
+/* Client-side API to finalize the PVAR profiling interface */
+hg_return_t HG_Prof_finalize() {
+
+  if(hg_prof_get_is_initialized()) {
+    hg_prof_set_is_initialized(0);
+  }
+
+  fprintf(stderr, "[MERCURY_PROF_INTERFACE] Successfully shutting down profiling interface\n");
+  return HG_SUCCESS;
+}
+
+/* Client-side API to retrieve the number of PVARs currently exported */
+int HG_Prof_pvar_get_num() {
+  if(hg_prof_get_is_initialized()) {
+    return NUM_PVARS;
+  } else {
+    return 0;
+  }
+}
+
+/* Gather information about every PVAR exported. This API is necessary in order for the client to discover the types, bindings, etc.
+ * The client can then allocate the necessary data structures using this information */
+hg_return_t HG_Prof_pvar_get_info(int pvar_index, char *name, int *name_len, hg_prof_class_t *var_class, hg_prof_datatype_t *datatype, char *desc, int *desc_len, hg_prof_bind_t *bind, int *continuous) {
+  
+  if(!hg_prof_get_is_initialized())
+    return HG_NA_ERROR;
+ 
+  assert(pvar_index < NUM_PVARS);
+ 
+  unsigned int key = pvar_index;
+  hg_prof_pvar_data_t * val;
+
+  /* Lookup the internal PVAR hash table to gather information about this PVAR */
+  val = hg_hash_table_lookup(pvar_table, (hg_hash_table_key_t)(&key));
+  strcpy(name, (*val).name);
+  *name_len = strlen(name);
+  strcpy(desc, (*val).description);
+  *desc_len = strlen(desc);
+  *var_class = (*val).pvar_class;
+  *datatype = (*val).pvar_datatype;
+  *bind = (*val).pvar_bind;
+  *continuous = (*val).continuous;
+
+  return HG_SUCCESS;
+}
+
+/* Create a session. In this case, return a reference to the default session that is currently supported */
+hg_return_t HG_Prof_pvar_session_create(hg_prof_pvar_session_t *session) {
+  if(!hg_prof_get_is_initialized())
+    return HG_NA_ERROR;
+
+  default_session.reference_counter += 1;
+
+  /* Only support one tool at the moment */
+  assert(default_session.reference_counter == 1);
+
+  *session = &default_session;
+
+  return HG_SUCCESS;
+}
+
+/* Allocate a handle for a PVAR at a given index.
+ * This handle will later be used by the client to query the value for the PVAR
+ * This handle is an opaque object */
+hg_return_t HG_Prof_pvar_handle_alloc(hg_prof_pvar_session_t session, int pvar_index, void *obj_handle, hg_prof_pvar_handle_t *handle, int *count) {
+
+  if(!hg_prof_get_is_initialized())
+    return HG_NA_ERROR;
+
+  /* Only supporting a default session and null bind type at the moment */
+  assert(session == &default_session);
+  assert(obj_handle == NULL);
+
+  struct hg_prof_pvar_session s = *session;
+  unsigned int key = pvar_index;
+  hg_prof_pvar_data_t * val;
+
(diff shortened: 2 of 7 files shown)


src/mercury_prof_interface.h
```diff
@@ -0,0 +1,37 @@
...
+ * found at the root of the source code distribution tree.
+ */
+
+#ifndef MERCURY_PROF_INTERFACE_H
+#define MERCURY_PROF_INTERFACE_H
+
+#include "mercury_prof_types.h"
+
+/*Initialize and finalize routines */
+HG_PUBLIC hg_return_t HG_Prof_init();
```
soumagne: Could you please add documentation for every function using the template that is similar to other mercury headers? i.e. something like:
```C
/**
 * <Description>
 *
 * \param param1 [IN/OUT]         <description>
 * \param param2 etc
 *
 * \return HG_SUCCESS or corresponding HG error code
 */
```
Since you added comments in the .c instead, you could just move those comments to the header. We usually don't duplicate function documentation between source and header so having it in one place in the header would be best.

srini009: @soumagne: I have addressed this issue in the latest commit.

soumagne: Great thanks! Yes that's much easier to review the API.

src/mercury_prof_interface.h
```diff
@@ -0,0 +1,126 @@
...
+#endif
+
+/**
+ * Client-side API to initialize the PVAR profiling interface.
+ *
+ * \return HG_SUCCESS or corresponding HG error code
+ */
+HG_PUBLIC hg_return_t 
+HG_Prof_init();
+
```
soumagne: That should probably take an hg_class as a parameter. We usually try to avoid using globals.

src/mercury_prof_interface.h
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Client-side API to finalize the PVAR profiling interface.
+ *
+ * \return HG_SUCCESS or corresponding HG error code
+ */
+HG_PUBLIC hg_return_t 
+HG_Prof_finalize();
+
```
soumagne: Same here

src/mercury_prof_interface.h
```diff
@@ -0,0 +1,126 @@
...
+ * \param obj_handle [IN]	  pointer to the object handle with which the PVAR is associated; can be NULL
+ * \param handle [IN/OUT]	  pointer to the opaque PVAR handle object that is returned to the calling client
+ * \param count [IN/OUT]	  pointer to the integer address representing the number of values associated with the PVAR
+ *
+ * \return HG_SUCCESS or corresponding HG error code
+ */
+HG_PUBLIC hg_return_t 
+HG_Prof_pvar_handle_alloc(hg_prof_pvar_session_t session, 
+			int pvar_index, void *obj_handle, hg_prof_pvar_handle_t *handle, int *count);
+
```
soumagne: if you have an alloc routine, you should also have a free routine I expect?

src/mercury_prof_interface.h
```diff
@@ -0,0 +1,126 @@
...
+HG_Prof_pvar_session_create(hg_prof_pvar_session_t *session);
+
+/**
+ * Client-side API to retrieve the number of PVARs currently exported.
+ *
+ * \return int representing the number of PVARs
+ */
+HG_PUBLIC int 
+HG_Prof_pvar_get_num();
+
```
soumagne: Also here

src/mercury_prof_interface.h
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Create a session. In this case, return a reference to the default session that is currently supported.
+ *
+ * \param session [IN/OUT]        pointer to PVAR session object 
+ *
+ * \return HG_SUCCESS or corresponding HG error code
+ */
+HG_PUBLIC hg_return_t 
+HG_Prof_pvar_session_create(hg_prof_pvar_session_t *session);
+
```
soumagne: Also here

src/mercury_prof_interface.h
```diff
@@ -0,0 +1,126 @@
...
+ * Start the PVAR is it is not continuous and has not been started yet.
+ *
+ * \param session [IN]        opaque PVAR session object 
+ * \param handle [IN]         opaque PVAR handle object 
+ *
+ * \return HG_SUCCESS or corresponding HG error code
+ */
+HG_PUBLIC hg_return_t 
+HG_Prof_pvar_start(hg_prof_pvar_session_t session, hg_prof_pvar_handle_t handle);
+
```
soumagne: Same here, I assume if you have a start you will need terminate or end?

src/mercury.c
```diff
@@ -22,6 +22,10 @@
 #include <string.h>
 #include <assert.h>
 
+/* PVAR profiling support */
+#include "mercury_prof_pvar_impl.h"
+HG_PROF_PVAR_UINT_COUNTER_DECL_EXTERN(hg_pvar_hg_forward_count);
+
```
soumagne: As a general comment I think it would be good to see if/how we can avoid declaring global variables.

carns:
I would also suggest that the comments for this interface reference a URL (I'm not sure which one off hand) for the MPI_T interface with a brief comment along the lines of "The following interface is patterned off of the MPI_T interface, except that X, Y, Z", where X, Y, an Z are notable philosophical differences (like being oriented around HG classes rather than globals) if there are any.

That way we can lean on the MPI documentation to some degree and also help rationalize the API conventions, since there is some value to reusing terminology for this stuff.

src/mercury_prof_pvar_impl.h
```diff
@@ -0,0 +1,89 @@
...
+        /* Set initial value */ \
+        PVAR_COUNTER_##name = 0; \
+        addr = &PVAR_COUNTER_##name; \
+        HG_PROF_PVAR_REGISTER_impl(HG_PVAR_CLASS_COUNTER, dtype, #name, \
+            addr, 1, bind, 1, desc); 
+
+/* Increment the value of a PVAR */
+#define HG_PROF_PVAR_COUNTER_INC(name, val) \
+    *(&PVAR_COUNTER_##name) += val;
+
```
soumagne: @srini009 watching your video, it made me think that you'll need to make sure that all increments are either done using atomics or are protected by a mutex. I think it makes sense to go with atomic types (see https://github.com/mercury-hpc/mercury/blob/master/src/util/mercury_atomic.h) rather than using locks.

carns: Do the performance counters actually need to be atomic?  I had assumed not, but I could be missing something.

soumagne: If you're counting the number of forward calls for example and increment the counter, this has to be protected by a lock or be done atomically.

carns: I definitely see that it could be updated (or accessed) concurrently, just wasn't sure about if that caused danger here or not.  But I guess "+= val" is technically a read/modify/write so you could lose updates if it's not atomic.

I'm on board; we should use the hg atomics.

srini009:
Hey guys @carns, @soumagne: Thank you for your useful comments. I generally agree with all the points that have been put forward. I shall address most if not all comments in the next 2 days. I hope that works!

srini009:
@soumagne, @carns: I have addressed all your comments. Kindly check if this is okay. I apologize for the delay, it took longer than I expected.

carns:
> @soumagne, @carns: I have addressed all your comments. Kindly check if this is okay. I apologize for the delay, it took longer than I expected.

Thanks @srini009 !

I just checked two things that I had commented on:

a) did you find a URL or some other reference to put in the comments for the general model the API is using?  It might be there and I'm missing it.

b) there has got to be a better way to implement COUNTER_INC() than looping over an incr() right?  Maybe do a get() and then a cas() instead (with a retry loop if needed)?  Jerome may have another idea.  I'm thinking of the case where the val could be relatively large, like if you are accumulating bytes transmitted or something.  The current code might be slow, and also isn't technically atomic across the macro.

srini009:
Hi @carns,
As for (a), I have addressed the comment by modifying the commit description (see top). Oops, I think you probably meant the code! :) I shall add the same comments in the interface header file as well.

As for (b), indeed, I did give this a thought. As far as atomicity is concerned, hopefully, I'm not mistaken, but I think we don't need to enforce atomicity at the macro level since individual updates are atomic and addition is commutative. I don't think we shall be encountering race conditions here. But you're right, it is incredibly inefficient for large updates related to bytes transferred, etc. Perhaps I can take @soumagne's help here. 

I am excited about this becoming a part of mercury. After that, it is a relatively quick step to introduce a bunch of new PVARs that are insightful.

Regards,

carns:
> Hi @carns,
> As for (a), I have addressed the comment by modifying the commit description (see top). Oops, I think you probably meant the code! :) I shall add the same comments in the interface header file as well.
> 
> As for (b), indeed, I did give this a thought. As far as atomicity is concerned, hopefully, I'm not mistaken, but I think we don't need to enforce atomicity at the macro level since individual updates are atomic and addition is commutative. I don't think we shall be encountering race conditions here. But you're right, it is incredibly inefficient for large updates related to bytes transferred, etc. Perhaps I can take @soumagne's help here.
> 
> I am excited about this becoming a part of mercury. After that, it is a relatively quick step to introduce a bunch of new PVARs that are insightful.
> 
> Regards,

Ah!  Yes, that link and explanation in the PR comment is perfect, just drop it in the code.  My thinking is that in the future someone might look at the code and want to change things without realizing that it's intentionally matching some broader conventions.

carns:
@soumagne what's the best what to do the performance counter macro?  Do a get(), calculate new value, then cas() (in a loop to retry if the cas fails)?

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+   int is_initialized;     /* Is profiling initialized */
+};
+
+/* HG class */
+struct hg_private_class {
+    struct hg_class hg_class;       /* Must remain as first field */
+    int hg_prof_is_initialized;     /* Is profiling initialized */
+    int num_pvars;          /* No of PVARs currently supported */
+    hg_prof_pvar_session_t session; /* Is profiling initialized on the session */
+};
```
soumagne, line 55: Maybe I am missing something but you should have a `struct hg_prof_class` here, not `hg_private_class`.

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+/*******************/
+/* Local Variables */
+/*******************/
+
+/*---------------------------------------------------------------------------*/
+static void 
+hg_prof_set_is_initialized(struct hg_private_class * hg_private_class)
+{
+  hg_private_class->hg_prof_is_initialized = 1;
+}
```
soumagne, line 66: Just a detail but instead of using 1 or 0, there are HG_TRUE and HG_FALSE macros

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+hg_prof_set_session_is_initialized(struct hg_prof_pvar_session * session)
+{
+  session->is_initialized = 1;
+}
+
+/*---------------------------------------------------------------------------*/
+static int 
+hg_prof_get_session_is_initialized(struct hg_prof_pvar_session * session) {
+   return session->is_initialized;
+}
```
soumagne, line 85: Same here, you could return an hg_bool_t instead.

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+/*---------------------------------------------------------------------------*/
+hg_return_t 
+HG_Prof_init(hg_class_t *hg_class) {
+
+  struct hg_private_class *hg_private_class = (struct hg_private_class *)(hg_class);
+
+  hg_prof_set_is_initialized(hg_private_class);
+  hg_private_class->num_pvars = NUM_PVARS;
+  hg_private_class->session = NULL;
+
```
soumagne, line 96: I don't think you are allowed to do that here :) HG_Prof_init() should return a new hg_prof_class_t * so I would expect you to malloc a new struct hg_prof_class in that routine, fill it and return it.

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+  return HG_SUCCESS;
+}
+
+/*---------------------------------------------------------------------------*/
+hg_return_t 
+HG_Prof_finalize(hg_class_t *hg_class) {
+
```
soumagne, line 103: Here instead it should take an hg_prof_class_t *

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  return HG_SUCCESS;
+}
+
+/*---------------------------------------------------------------------------*/
+int 
+HG_Prof_pvar_get_num(hg_class_t *hg_class) {
+  struct hg_private_class *hg_private_class = (struct hg_private_class *)(hg_class);
+
```
soumagne, line 118: Same here and following routines

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+  if(hg_prof_get_is_initialized(hg_private_class)) {
+    return hg_private_class->num_pvars;
+  } else {
+    return 0;
+  }
+}
+
+/*---------------------------------------------------------------------------*/
+hg_return_t 
+HG_Prof_pvar_get_info(hg_class_t *hg_class, int pvar_index, char *name, int *name_len, hg_prof_class_t *var_class, hg_prof_datatype_t *datatype, char *desc, int *desc_len, hg_prof_bind_t *bind, int *continuous) {
```
soumagne, line 128: So that means hg_prof_class_t becomes hg_prof_var_class_t ?

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+
+  if(!hg_prof_get_session_is_initialized(session))
+    return HG_NA_ERROR;
+
+  /* Only supporting a null bind type at the moment */
+  assert(obj_handle == NULL);
+
+  struct hg_prof_pvar_session s = *session;
+  unsigned int key = pvar_index;
+  hg_prof_pvar_data_t * val;
```
soumagne, line 199: Please keep declarations at beginning of blocks

src/mercury_prof_interface.c
```diff
@@ -0,0 +1,281 @@
...
+
+  return HG_SUCCESS;
+}
+
+/*---------------------------------------------------------------------------*/
+hg_return_t 
+HG_Prof_pvar_handle_free(hg_prof_pvar_session_t session, int pvar_index, hg_prof_pvar_handle_t *handle) {
+
+  if(!hg_prof_get_session_is_initialized(session))
+    return HG_NA_ERROR;
```
soumagne, line 228: that seems like the wrong error code :) HG_NA_ERROR means it's an NA layer error so maybe just HG_INVALID_ARG?

src/mercury_prof_interface.h
soumagne: See my previous comment, you should return an `hg_prof_class_t *` here

src/mercury_prof_types.h
```diff
@@ -0,0 +1,42 @@
...
+
+/* Enumerate the various types of PVAR classes */
+typedef enum {
+   HG_PVAR_CLASS_STATE, /* PVAR that represents any one of a set of discrete states */
+   HG_PVAR_CLASS_COUNTER, /* PVAR that represents a regular monotonic counter */
+   HG_PVAR_CLASS_LEVEL, /* PVAR that represents a utilization level (in percentage) of a given resource */
+   HG_PVAR_CLASS_SIZE, /* PVAR that represents the size of a given resource at any given point in time */
+   HG_PVAR_CLASS_HIGHWATERMARK, /* PVAR that represents a high watermark value */
+   HG_PVAR_CLASS_LOWWATERMARK /* PVAR that represents a low watermark value */
+} hg_prof_class_t;
```
soumagne, line 34: yes you should rename that one to hg_prof_var_class_t, given the changes that I suggested earlier.

soumagne:
@carns @srini009 yes for the performance instead of:
```C
#define HG_PROF_PVAR_COUNTER_INC(name, val)
    addr_##name = (addr_##name == NULL ? hg_prof_get_pvar_addr_from_name(#name): addr_##name);
    for(int i=0; i < val; i++)
        hg_atomic_incr32(addr_##name);
```
you should do something like:
```C
#define HG_PROF_PVAR_COUNTER_INC(name, val) do {
    hg_util_int32_t tmp;

    addr_##name = (addr_##name == NULL ? hg_prof_get_pvar_addr_from_name(#name): addr_##name);
    do {
        tmp = hg_atomic_get32(addr_##name);
    } while (!hg_atomic_cas32(addr_##name, tmp, (tmp + val)));
} while (0)
```

PR #361, opened 2020-05-14
Addr_lookup and SM changes

soumagne:


src/mercury.h
```diff
@@ -574,18 +574,38 @@ HG_Registered_disabled_response(
...
 HG_PUBLIC hg_return_t
-HG_Addr_lookup(
+HG_Addr_lookup1(
         hg_context_t *context,
         hg_cb_t       callback,
         void         *arg,
         const char   *name,
         hg_op_id_t   *op_id
         );
+#define HG_Addr_lookup HG_Addr_lookup1
```
carns, line 589: What is the transition plan when people want to use the newer style lookup?  They shoudl use HG_Addr_lookup2() explicitly?

I'm not sure how to handle this, but maybe users could #define something before the inclusion of mercury.h to toggle which is the default?

soumagne: Yes you're right it's maybe best not to try to take decisions for users, I'll remove the define. I was looking at that too: https://developers.redhat.com/blog/2019/08/01/how-the-gnu-c-library-handles-backward-compatibility/ maybe that's the right way of doing it but it is maybe overkill in that case.

carns: That glibc strategy is interesting, but yeah, I would agree, probably overkill for this use case :)

soumagne: I removed the define in the header.

carns: I feel like I'm bike-shed painting now, but I can't help it :)

Maybe just an `#ifndef HG_Addr_lookup \ #define HG_Addr_lookup HG_Addr_lookup1 \ #endif`?  That way legacy codes will continue to work fine with the upcoming release, but people that want to switch over can either #define HG_Addr_lookup before including Mercury, or else use 1 or 2 explicitly?

soumagne: No problem, it's the time to make changes :)

That seems fine with me, and does not hurt anyone this way. OK will update it after I fixed the CI.

PR #367, opened 2020-06-17
WIP: include -Wl,-rpath in pkgconfig Libs:

carns:
Marked as a WIP for discussion.  Rationale: Mercury is usually installed in non-system paths (either by HPC system administrator or HPC system user).  Executables that user Mercury need to be able to find this library path at runtime, which generally means using LD_LIBRARY_PATH or embedded rpaths.  LD_LIBRARY_PATH isn't a great universal solution because it relies on being able to propagate an environment variable wherever the executable is launched.  RPATH/RUNPATH/NEEDED is conceptually easier, but different compilers have different default settings for how they handle library dependencies, making it unpredictable if a given system will automatically get it right.

This patch adds an explicit rpath in the mercury.pc file to make sure that external users set the embedded rpath correctly regardless of compiler/binutils default settings.

Marked as WIP because I'm not entirely certain this is the sensible thing to do.  If this doesn't look generally correct, we could modify mercury.pc in the spack package instead.  Or if it is the wrong approach altogether we can find a way to tweak margo to assume an rpath for Mercury, but I'm not sure if other Mercury users could hit this problem.

Confirmed this produces Margo/SSG executables on Summit with correct paths for Mercury.

@chuckcranor @roblatham00


carns:
I'm trying some other options too, will report back soon.

carns:
Closing this PR; we have a workaround in the margo build system that's less invasive.  Will land that one instead later today.

PR #374, opened 2020-07-07
util: refine hg_mem_get_page_size

liuxuezhao:
Avoid querying pagesize every time

Signed-off-by: Johann Lombardi <johann.lombardi@intel.com>


liuxuezhao:
@soumagne  the CDash failure is some memleaks in ofi, such as https://mercury-cdash.hdfgroup.org/testDetails.php?test=15303&build=471 . should be unrelated with this patch. thanks.

PR #384, opened 2020-08-12
bulk: fix an bug in hg_bulk_transfer_cb

liuxuezhao:
When the bulk transfer failed, should increase op_completed_count
and complete the bulk when needed, or the bulk possibly cannot complete
forever.


soumagne:
Will merge accompanying test and further clean up in following PR.

PR #430, opened 2021-02-04
HG util: remove check for STDERR_FILENO (fix #428)

### LINKED ISSUE #428: daemonization causes na+sm initialization failure
**Describe the bug**
UnifyFS has an option to daemonize its server, which makes the server call its `daemonize()` function before initializing margo/mercury. When this option is enabled, `margo_init("na+sm", ...)` fails.

**To Reproduce**
I created a simple reproducer (see below) based on the example mercury server in the documentation (https://mochi.readthedocs.io/en/latest/mercury/01_init.html). The failure can be reproduced when this code is compiled with `-DDAEMONIZE=1` and run passing "na+sm" as the first parameter. Without that flag defined, the example works.

**Platform (please complete the following information):**
 - Ubuntu 18.04 docker container (based on spack/ubuntu-bionic) with `kernel.yama.ptrace_scope=0`
 - Compiler version: `gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0`
 - Spack spec: `mercury@2.0.0 ^libfabric fabrics=rxm,sockets,tcp`
 - Plugin and protocol used: `na+sm`

**Reproducer Code**
```
#include <assert.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>
#include <unistd.h>
#include <mercury.h>

static hg_class_t*     hg_class   = NULL; /* the mercury class */
static hg_context_t*   hg_context = NULL; /* the mercury context */

/*
 * Perform steps to create a daemon process:
 *
 *  1. Fork and exit from parent so child runs in the background
 *  2. Set the daemon umask to 0 so file modes passed to open() and
 *     mkdir() fully control access modes
 *  3. Call setsid() to create a new session and detach from controlling tty
 *  4. Change current working directory to / so daemon doesn't block
 *     filesystem unmounts
 *  5. close STDIN, STDOUT, and STDERR
 *  6. Fork again to abdicate session leader position to guarantee
 *     daemon cannot reacquire a controlling TTY
 *
 */
static void daemonize(void)
{
    pid_t pid;
    pid_t sid;
    int rc;

    pid = fork();
    if (pid < 0) {
        fprintf(stderr, "first fork failed: %s", strerror(errno));
        exit(1);
    }

    if (pid > 0) {
        exit(0);
    }

    umask(0);

    sid = setsid();
    if (sid < 0) {
        fprintf(stderr, "setsid failed: %s", strerror(errno));
        exit(1);
    }

    rc = chdir("/");
    if (rc < 0) {
        fprintf(stderr, "chdir failed: %s", strerror(errno));
        exit(1);
    }

    close(STDIN_FILENO);
    close(STDOUT_FILENO);
    close(STDERR_FILENO);

    pid = fork();
    if (pid < 0) {
        assert(!"second fork failed");
        exit(1);
    } else if (pid > 0) {
        exit(0);
    }
}

int main(int argc, char** argv)
{
    hg_return_t ret;
	char* proto = argv[1];

#ifdef DAEMONIZE
    daemonize();
#endif

    FILE* log = fopen("/tmp/test_mercury_server.log", "w");

    /* Initialize Mercury and get an hg_class handle.
     * bmi+tcp is the protocol to use.
     * localhost is the address of the server (not useful at the server itself).
     * HG_TRUE is here to specify that mercury will listen for incoming requests.
     * (HG_TRUE on servers, HG_FALSE on clients).
     */
    hg_class = HG_Init(proto, HG_TRUE);
    if (hg_class == NULL) {
        if (NULL != log) {
            fprintf(log, "HG_Init(%s, HG_TRUE) failed\n", proto);
            fflush(log);
            fclose(log);
        }
        exit(1);
    }

    /* Get the address of the server */
    char hostname[128];
    memset(hostname, 0, sizeof(hostname));
    hg_size_t hostname_size = sizeof(hostname);
    hg_addr_t self_addr;
    HG_Addr_self(hg_class, &self_addr);
    HG_Addr_to_string(hg_class, hostname, &hostname_size, self_addr);
    if (NULL != log) {
        fprintf(log, "Server running at address: %s\n", hostname);
    }
    HG_Addr_free(hg_class, self_addr);

    /* Creates a Mercury context from the Mercury class. */
    hg_context = HG_Context_create(hg_class);
    if (hg_context == NULL) {
        if (NULL != log) {
            fprintf(log, "HG_Context_create() failed\n");
        }
    } else {
        /* Destroys the Mercury context. */
        ret = HG_Context_destroy(hg_context);
        if (ret != HG_SUCCESS) {
            if (NULL != log) {
                fprintf(log, "HG_Context_destroy() failed\n");
            }
        }
    }

    /* Finalize Mercury. */
    ret = HG_Finalize(hg_class);
    if (ret != HG_SUCCESS) {
        if (NULL != log) {
            fprintf(log, "HG_Finalize() failed\n");
        }
    }

    if (NULL != log) {
        fclose(log);
    }

    return 0;
}
```

soumagne:
Returning an error when adding a file descriptor lower than STDERR_FILENO
was causing errors when users were closing those std file descriptors

### DIFF
--- a/src/mercury_core.c
+++ b/src/mercury_core.c
@@ -1127,7 +1127,7 @@ hg_core_context_create(hg_core_class_t *hg_core_class, hg_uint8_t id,
         event.data.u32 = (hg_util_uint32_t) HG_CORE_POLL_NA;
         rc = hg_poll_add(context->poll_set, na_poll_fd, &event);
         HG_CHECK_ERROR(rc != HG_UTIL_SUCCESS, error, ret, HG_NOMEM,
-            "hg_poll_add() failed");
+            "hg_poll_add() failed (na_poll_fd=%d)", na_poll_fd);
 
 #ifdef NA_HAS_SM
         if (context->core_context.na_sm_context) {
@@ -1139,7 +1139,7 @@ hg_core_context_create(hg_core_class_t *hg_core_class, hg_uint8_t id,
             event.data.u32 = (hg_util_uint32_t) HG_CORE_POLL_SM;
             rc = hg_poll_add(context->poll_set, na_poll_fd, &event);
             HG_CHECK_ERROR(rc != HG_UTIL_SUCCESS, error, ret, HG_NOMEM,
-                "hg_poll_add() failed");
+                "hg_poll_add() failed (na_poll_fd=%d)", na_poll_fd);
         }
 #endif
 
@@ -1154,7 +1154,7 @@ hg_core_context_create(hg_core_class_t *hg_core_class, hg_uint8_t id,
             rc = hg_poll_add(
                 context->poll_set, context->completion_queue_notify, &event);
             HG_CHECK_ERROR(rc != HG_UTIL_SUCCESS, error, ret, HG_NOMEM,
-                "hg_poll_add() failed");
+                "hg_poll_add() failed (na_poll_fd=%d)", na_poll_fd);
         }
     }
 
--- a/src/util/mercury_poll.c
+++ b/src/util/mercury_poll.c
@@ -193,9 +193,6 @@ hg_poll_add(hg_poll_set_t *poll_set, int fd, struct hg_poll_event *event)
 #endif
     int ret = HG_UTIL_SUCCESS;
 
-    HG_UTIL_CHECK_ERROR(fd <= STDERR_FILENO, done, ret, HG_UTIL_FAIL,
-        "fd is not valid (%d)", fd);
-
 #if defined(_WIN32)
     /* TODO */
 #elif defined(HG_UTIL_HAS_SYSEPOLL_H)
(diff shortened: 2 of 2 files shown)


soumagne:
My original intent was to prevent from adding those file descriptors by mistake to the polling sets but it looks like that's not compatible with that use case. Just removing that check for now, any better idea?

MichaelBrim:
@soumagne I don't know if it's a better idea, but what we did in MRNet to avoid the stdio file descriptor numbers in our poll set was to keep calling socket() (or whatever syscall created the fd) until the result was above STDERR_FILENO. We also left open the stdio fds to ensure future operations wouldn't use them.

gnuoyd:
This looks fine.

In *BSD a library function, `daemon(3)`, daemonizes a process. It detaches descriptors [0, 2] by performing `fd = open("/dev/null", O_RDWR)` and then `dup2(fd, STDIN_FILENO)`, `dup2(fd, STDOUT_FILENO)`, `dup2(fd, STDERR_FILENO)`. Sounds like UnifyFS should have cribbed that function. :-)

gnuoyd (approved):

carns:
I know I'm late, but I agree with this patch.  We hit the same issue in PVFS. Once a file descriptor has been closed (including stdin, stdout, and stderr) those FD numbers are fair game for Linux to reuse for any purpose.  The only ones you can trust to be truly invalidate are negative numbers.

PR #432, opened 2021-02-09
NA OFI: yield to other threads when using PSM2

soumagne:
When not yielding, a thread dedicated to progress would busy spin and not allow for other threads to be scheduled.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -71,6 +71,7 @@
 #include <inttypes.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <sched.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


carns (approved):
Looks great.  I tested this on the Bebop machine at ANL just now.  I'm running @marcvef 's test case, with 1 multithreaded server and 16 clients issuing concurrent RPCs.  This PR improved the RPC rate in this example from 557 ops/s to 58,355 ops/s.

chuckcranor:
Simple enough and only impacts ofi+psm2, so ok with me....   i'm still not entirely clear on what the spin is in marcvef's test case though.  does this fix the underlying problem, or just mask it?

chuckcranor (approved):

carns:
> Simple enough and only impacts ofi+psm2, so ok with me.... i'm still not entirely clear on what the spin is in marcvef's test case though. does this fix the underlying problem, or just mask it?

My understanding of the scenario is this: one thread is spinning on HG_Progress() (occasionally alternating to HG_Trigger()) while another pool of threads calls HG_Respond().  Something in the HG_Respond() path, probably down in libfabric or opa-psm2, is unable to acquire a lock because the progress thread is dominating it.

So this PR is a "fix" in that it artificially introduces some fairness in the locking by letting the thread pool get a crack at the lock if the progress thread drops the locks without having found any work to do.  I'm not thrilled with have to use a yield for this purpose, but the scope of the change is pretty limited.

PR #433, opened 2021-03-11
hg_request_wait(): simplify timeout logic & arithmetic

gnuoyd:
The previous countdown to timeout was hard to follow, and I wasn't sure about using a double to keep the time remaining.

This change compares the current time with an hg_time_t deadline to see if it's
time to exit. Should still run through the loop at least once if the timeout is 0 milliseconds.

I added a new utility function, hg_time_to_ms(), for converting an hg_time_t directly
to integer milliseconds.

### DIFF
--- a/src/util/mercury_request.c
+++ b/src/util/mercury_request.c
@@ -105,20 +105,19 @@ hg_request_destroy(hg_request_t *request)
 
 /*---------------------------------------------------------------------------*/
 int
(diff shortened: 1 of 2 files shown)


soumagne:
Thanks for adding that, I think it looks better that way. Let me fix the CI though, some weirdness happened there.

PR #436, opened 2021-03-23
Improve logging and add subsystems, debug traces and report

soumagne:
This adds a new environment variable `HG_LOG_SUBSYS` as well as multiple log subsystems. Log subsystems that depend on a given parent log will be automatically turned on unless they are too verbose, in which case they can manually enabled.
For instance: `HG_LOG_SUBSYS=na` will turn on NA logs. `HG_LOG_SUBSYS=op,progress` will turn on op and progress related logs. `HG_LOG_SUBSYS=hg` will turn on HG logs as well as NA logs (HG log is parent of NA log). It is also possible to prevent a subsystem from being enabled, e.g., `HG_LOG_SUBSYS=hg,~na` will only print HG logs and turn off NA logs.

Additionally a new log level has been added, `HG_LOG_LEVEL=min_debug`. This level prevents log output from being printed but records a trace and counters that will be printed when an error occurs on a given subsystem. Here is an example:
```
# [99070.800103] fatal: [error] /home/jsoumagne/workspace/mercury-hpc/mercury/src/na/na_ofi.c:1618 
## na_ofi_provider_check(): Requested OFI provider "gni" (derived from "gni"
   protocol) is not available. Please re-compile libfabric with support for
   "gni" or use one of the following available providers:
   tcp;ofi_rxm,udp;ofi_rxd,shm,udp,tcp,sockets
# [99070.800121] na: [error] /home/jsoumagne/workspace/mercury-hpc/mercury/src/na/na_ofi.c:1721 
## na_ofi_getinfo(): fi_getinfo() failed, rc: -61 (No data available)
### ----------------------
### (na) debug log summary
### ----------------------
# COUNTERS
# -
# NLOGS ENTRIES: 1
# [99070.684087] /home/jsoumagne/workspace/mercury-hpc/mercury/src/na/na.c:265
## NA_Initialize_opt()
```


carns:
Is there a runtime API for changing the subsys, or just the level?  (I see the `HG_Set_log_level()` function).

carns:
Overall this looks great.  My only other comment is that it might be helpful if the prefix on the messages (like `[99070.800103] fatal: [error] ` in the example given above also indicates explicitly that that it is a mercury log message.

I just tested this PR out with a Mochi service example that was using it's own logging too, and I could imagine an end user not realizing which underlying software component had emitted the error.

soumagne:
@carns yes there are `hg_log_set_level()` and `hg_log_set_subsys()`, those are in mercury log and not directly exposed through the HG API. `HG_Set_log_level()` actually does two things now, it set the log level to what you ask for and also enables the `hg` subsys module.

soumagne:
you're also right, I'll add an extra label that mentions the mercury package name.

carns:
> @carns yes there are `hg_log_set_level()` and `hg_log_set_subsys()`, those are in mercury log and not directly exposed through the HG API. `HG_Set_log_level()` actually does two things now, it set the log level to what you ask for and also enables the `hg` subsys module.

Seems like it the public API should let you change both?  I'm just thinking about it from the perspective of services layered atop Mercury that might want to be able to control logging.

soumagne:
@carns I addressed both comments that you had and also added an HG_Set_log_subsys() API call.

carns:
Thanks!

PR #444, opened 2021-04-06
NA: add max_unexpected_size and max_expected_size hints (fix #263)

### LINKED ISSUE #263: Allow hint to request large buffer size
*** Overview

Allow users to suggest new buffer sizes, including the default unexpected message size and the expected message size, during HG_Init_opt() with an hg_init_info struct. If the suggested value is incompatible with the selected native transport, mercury ignores the suggestion.

In addition, there could be an API to allow users to query NA capabilities so the user knows the max possible buffer sizes he/she can set.

*** Details

Mercury has 2 types of messages: normal RPC and bulk transfer. Normal RPCs can only transfer data within a buffer size threshold while bulk transfer has virtually no limits on data sizes.

If an RPC message size is larger than the buffer threshold mercury will automatically convert it  to a bulk transfer. This is done by first re-allocating enough buffer memory (realloc) to hold the message data, and then performing a bulk transfer to transfer it.
 
Different NA has different such buffer threshold.

For mercury applications (e.g., DeltaFS) that are network bandwidth bound and only send very large messages (e.g., 32KiB), each such message is either sent as a direct bulk transfer, or done as a normal RPC but implicitly converted to a bulk transfer. Either way, there will be 2 round trips per message: 1 for sending the bulk transfer descriptor, and the other for transmitting of the actual data.

If we allow a mercury application to dynamically suggest a larger buffer size, the application would then be able to send large data using just a normal RPC. This reduces the RPC cost to 1 round trip per message and can improve performance sometimes significantly.

Here's a quick test using CMU's Narwhal machine.

First, we hack the na_bmi.c, increasing the default unexpected message size from 4K to 10k, and then reducing the expected message size from 4K to only 64 bytes.

1 diff --git a/src/na/na_bmi.c b/src/na/na_bmi.c
2 index 042debc..9bd9267 100644
3 --- a/src/na/na_bmi.c
4 +++ b/src/na/na_bmi.c
5 @@ -35,8 +35,8 @@
6 #define NA_BMI_DEFAULT_PORT_TRIES 128
7
8 /* Msg sizes */
9 -#define NA_BMI_UNEXPECTED_SIZE 4096
10 -#define NA_BMI_EXPECTED_SIZE NA_BMI_UNEXPECTED_SIZE
11 +#define NA_BMI_UNEXPECTED_SIZE 10000
12 +#define NA_BMI_EXPECTED_SIZE 64
13
14 /* Max tag */
15 #define NA_BMI_MAX_TAG (NA_TAG_UB >> 2)
 

Here's the results of running mercury-runner (1 sender process, 1 receiver process, 2 nodes) on CMU's Narwhal machine (1Gb Ethernet) using 8K rpc sizes:
 
== Before the hack using 1 to 16 in-flight RPCs (each rpc is a normal RPC + a bulk transfer)
 
bmi 1 0.000567 sec/op
bmi 2 0.000337 sec/op
bmi 4 0.000187 sec/op
bmi 8 0.000143 sec/op
bmi 16 0.000140 sec/op
 
== After the hack (each rpc is just a normal RPC)
 
bmi 1 0.000374 sec/op
bmi 2 0.000212 sec/op
bmi 4 0.000118 sec/op
bmi 8 0.000086 sec/op
bmi 16 0.000075 sec/op
 
The overall throughput is improved by about 1.5-1.8x.

soumagne:
NA OFI: add support for msg size hints and refactor domain open code

Add -Z na test option to control max msg size

### DIFF
--- a/Testing/na/na_test.c
+++ b/Testing/na/na_test.c
@@ -161,6 +161,9 @@ na_test_parse_options(int argc, char *argv[], struct na_test_info *na_test_info)
                 na_test_info->max_contexts =
                     (na_uint8_t) atoi(na_test_opt_arg_g);
                 break;
+            case 'Z': /* msg size */
+                na_test_info->max_msg_size = atoi(na_test_opt_arg_g);
+                break;
             case 'V': /* verbose */
                 na_test_info->verbose = NA_TRUE;
                 break;
@@ -427,6 +430,8 @@ NA_Test_init(int argc, char *argv[], struct na_test_info *na_test_info)
     }
     na_init_info.auth_key = na_test_info->key;
     na_init_info.max_contexts = na_test_info->max_contexts;
+    na_init_info.max_unexpected_size = (na_size_t) na_test_info->max_msg_size;
+    na_init_info.max_expected_size = (na_size_t) na_test_info->max_msg_size;
 
     printf("# Using info string: %s\n", info_string);
     na_test_info->na_class =
--- a/src/na/na_types.h
+++ b/src/na/na_types.h
@@ -31,10 +31,12 @@ typedef na_uint64_t na_offset_t;               /* Offset */
 
 /* Init info */
 struct na_init_info {
-    const char *ip_subnet;     /* Preferred IP subnet */
-    const char *auth_key;      /* Authorization key */
-    na_uint32_t progress_mode; /* Progress mode */
-    na_uint8_t max_contexts;   /* Max contexts */
+    const char *ip_subnet;         /* Preferred IP subnet */
+    const char *auth_key;          /* Authorization key */
+    na_size_t max_unexpected_size; /* Max unexpected size hint */
+    na_size_t max_expected_size;   /* Max expected size hint */
+    na_uint32_t progress_mode;     /* Progress mode */
+    na_uint8_t max_contexts;       /* Max contexts */
 };
 
 /* Segment */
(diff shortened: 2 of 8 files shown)


src/na/na_types.h
```diff
@@ -31,10 +31,12 @@ typedef na_uint64_t na_offset_t;               /* Offset */
...
 /* Init info */
 struct na_init_info {
-    const char *ip_subnet;     /* Preferred IP subnet */
-    const char *auth_key;      /* Authorization key */
-    na_uint32_t progress_mode; /* Progress mode */
-    na_uint8_t max_contexts;   /* Max contexts */
+    const char *ip_subnet;         /* Preferred IP subnet */
+    const char *auth_key;          /* Authorization key */
+    na_size_t max_unexpected_size; /* Max unexpected size hint */
+    na_size_t max_expected_size;   /* Max expected size hint */
```
soumagne, line 37: Added those 2 new flags. These are just hints provided by users and NA plugins do not have to honor those limits if they can't / don't want to support it.

Testing/na/na_test.c
```diff
@@ -161,6 +161,9 @@ na_test_parse_options(int argc, char *argv[], struct na_test_info *na_test_info)
                 na_test_info->max_contexts =
                     (na_uint8_t) atoi(na_test_opt_arg_g);
                 break;
+            case 'Z': /* msg size */
+                na_test_info->max_msg_size = atoi(na_test_opt_arg_g);
+                break;
```
soumagne, line 166: @gnuoyd also added a -Z option to run the NA tests. Just thinking that you may want to use it with UCX too.

Issue #445, opened 2021-04-12
NA SM: to lookup address which isn't exists in the hash table makes memory leak


KoyamaSohei:
**Describe the bug**
In `na+sm`, when mercury try to lookup address which isn't exists in the hash table, mercury insert that into hash table. 

https://github.com/mercury-hpc/mercury/blob/4325efabcdaf5d0a8847015aac7ab9c168f9b38b/src/na/na_sm.c#L3822

At this time, the reference counter is initialized to 1 in `na_sm_addr_create`

https://github.com/mercury-hpc/mercury/blob/4325efabcdaf5d0a8847015aac7ab9c168f9b38b/src/na/na_sm.c#L2381

However, in `na_sm_addr_lookup`, increment reference counter, so `&na_sm_addr->ref_count` becomes 2.

https://github.com/mercury-hpc/mercury/blob/4325efabcdaf5d0a8847015aac7ab9c168f9b38b/src/na/na_sm.c#L3836

After that, in `na_sm_addr_free`, the reference counter is decremented, not released yet because it is not 0.

https://github.com/mercury-hpc/mercury/blob/4325efabcdaf5d0a8847015aac7ab9c168f9b38b/src/na/na_sm.c#L3855

Eventually the address will not be released and will result in a memory leak.

**To Reproduce**

- `"na+sm://4661/0" ` is dummy address
- use [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) to detect memory leak
```
$ cat main.c
#include <mercury.h>
#include <assert.h>

int main() {
  na_class_t *class;
  na_addr_t addr = NA_ADDR_NULL;

  NA_Set_log_level("debug");

  class = NA_Initialize("na+sm",NA_FALSE);
  assert(class != NULL);

  assert(NA_Addr_lookup(class,"na+sm://4661/0",&addr)==NA_SUCCESS);

  assert(NA_Addr_free(class,addr)==NA_SUCCESS);
  assert(NA_Finalize(class)==NA_SUCCESS);
}
$ gcc -Wall -g `pkg-config --cflags mercury` -fsanitize=address -fno-omit-frame-pointer main.c `pkg-config --libs mercury` -fsanitize=address
$ ASAN_OPTIONS="abort_on_error=1" ./a.out 
...
==4735==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 72 byte(s) in 1 object(s) allocated from:
[...]

PR #446, opened 2021-04-14
na sm: check if entry is in poll list

KoyamaSohei:
### Description

on `na_sm_addr_free`, there is a possibility that `na_sm_addr` is not in  list of addresses to poll.

For example, like the code of https://github.com/mercury-hpc/mercury/issues/445 , it can happen if we call `NA_Addr_free` without doing forwarding.

### Additional context

Currently due to issue https://github.com/mercury-hpc/mercury/issues/445 , these processes are not performed because the refcount is always 1 or more.

However, after fixing https://github.com/mercury-hpc/mercury/issues/445 , this case can occur.

### Additional context 2

I Implemented this with reference to `na_ofi_domain_close` in `na_ofi.c`

https://github.com/mercury-hpc/mercury/blob/4325efabcdaf5d0a8847015aac7ab9c168f9b38b/src/na/na_ofi.c#L2311

### DIFF
--- a/src/na/na_sm.c
+++ b/src/na/na_sm.c
@@ -3861,7 +3861,8 @@ na_sm_addr_free(na_class_t *na_class, na_addr_t addr)
 
     /* Remove address from list of addresses to poll */
     hg_thread_spin_lock(&na_sm_endpoint->poll_addr_list.lock);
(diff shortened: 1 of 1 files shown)


soumagne:
Fixed in #448

PR #477, opened 2021-06-11
Tidy up UCX thread configuration

soumagne:


src/na/na_ucx.c
```diff
@@ -924,6 +985,18 @@ na_ucx_initialize(na_class_t *nacl, const struct na_info NA_UNUSED *na_info,
...
+        NA_LOG_ERROR("context attributes contain no thread mode");
+        ret = NA_PROTOCOL_ERROR;
+        goto cleanup;
+    }
+
+    /* Best effort to ensure thread safety */
+    if (context_thread_mode != UCS_THREAD_MODE_SINGLE
+        && uctx_attrs.thread_mode < context_thread_mode)
+        NA_LOG_WARNING("Context thread mode is: %s",
+            ucs_thread_mode_names[uctx_attrs.thread_mode]);
```
gnuoyd, line 998: Two-line statement -> open & close curly braces?

src/na/na_ucx.c
```diff
@@ -860,6 +881,45 @@ na_ucx_initialize(na_class_t *nacl, const struct na_info NA_UNUSED *na_info,
...
+    }
+
+    /* Best effort to ensure thread safety */
+    if (worker_thread_mode != UCS_THREAD_MODE_SINGLE
+        && ucp_lib_attrs.max_thread_level < worker_thread_mode) {
+        worker_thread_mode = ucp_lib_attrs.max_thread_level;
+        NA_LOG_WARNING("Max worker thread level is: %s",
+            ucs_thread_mode_names[worker_thread_mode]);
+    }
+#endif
```
gnuoyd, line 921: The plugin should honor the thread mode that the caller requests or else fail with a meaningful error return.

src/na/na_types.h
```diff
@@ -135,10 +136,18 @@ typedef int (*na_cb_t)(const struct na_cb_info *callback_info);
...
 #define NA_NO_RETRY 0x02 /*!< no retry of operations in progress */
 
+/* Thread modes (default is thread-safe) */
+#define NA_THREAD_MODE_SINGLE_CLS                                              \
+    (0x01) /*!< only one thread will access class */
+#define NA_THREAD_MODE_SINGLE_CTX                                              \
+    (0x02) /*!< only one thread will access context */
+#define NA_THREAD_MODE_SINGLE                                                  \
+    (NA_THREAD_MODE_SINGLE_CLS | NA_THREAD_MODE_SINGLE_CTX)
+
```
gnuoyd, line 146: It seems like the number of threads is not the important thing. Rather, it's important that if more than one thread uses the same class or context, that their access is serialized? So _SERIAL rather than _SINGLE here?

soumagne: Yes that's true, I think I like that change to serial

src/na/na_ucx.c
soumagne: ok yes was trying to be a little permissive but probably better to fail, will fix

src/na/na_ucx.c
```diff
@@ -860,6 +881,45 @@ na_ucx_initialize(na_class_t *nacl, const struct na_info NA_UNUSED *na_info,
...
+
```
gnuoyd, line 922: I don't think the conditionally-compiled code really adds anything.

src/na/CMakeLists.txt
```diff
@@ -171,6 +173,17 @@ if(NA_USE_UCX)
...
     ${UCX_LIBRARIES}
   )
+  # Additional cmake testing
+  set(CMAKE_REQUIRED_INCLUDES ${UCX_INCLUDE_DIR})
+  set(CMAKE_REQUIRED_LIBRARIES ${UCX_LIBRARIES})
+
+  # Detect ucp_lib_query
+  check_symbol_exists(ucp_lib_query ucp/api/ucp.h NA_UCX_HAS_LIB_QUERY)
+  # Detect whether UCX has thread mode names
+  check_variable_exists(ucs_thread_mode_names NA_UCX_HAS_THREAD_MODE_NAMES)
```
gnuoyd, line 183: Is `ucs_thread_mode_names` a documented part of the API? Something new since 1.10?

src/na/na_ucx.c
soumagne: it's technically still one line :) but anw this one should probably also fail in the same fashion so I'll add that too

src/na/na_ucx.c
soumagne: it actually does but I should have added some explanation. If UCX is compiled without mt, the thread level returned by that routine will be UCS_THREAD_MODE_SERIALIZED and not UCS_THREAD_MODE_SINGLE. We can't really guess that without it and it allows to pass that down to the context code. If you pass UCS_THREAD_MODE_MULTI and UCX is compiled without mt, it would return 0 and not even try to use UCS_THREAD_MODE_SERIALIZED.

soumagne: And at the very least it also adds some confirmation on what the library supports.

src/na/CMakeLists.txt
soumagne: it's in master yes

PR #508, opened 2021-09-26
Change to na_ucx.c parsing of address string.  

jgmoore-or:
Allow specification device in the string.

Signed-off-by: Joseph Moore <joseph.moore@intel.com>


soumagne:
Thanks @jgmoore-or I will integrate it while updating the UCX plugin.

PR #545, opened 2022-02-14
NA OFI: two changes to FI_EAGAIN/retry handling

liuxuezhao:
1. When get FI_ENOTCONN/FI_ECONNRESET/FI_ECONNABORTED/FI_ECONNREFUSED for
    one OP, check the per context retry_op_queue to abort all ops to the
    same target fi_addr, to avoid retry that op.
2. Set a timestamp for first time retry, if following retry keep get
     EAGAIN within NA_OFI_OP_RETRY_TIMEOUT (60 second), give up the retry.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -231,6 +231,9 @@ static unsigned long const na_ofi_prov_flags[] = {NA_OFI_PROV_TYPES};
 #define NA_OFI_OP_QUEUED    (1 << 3)
 #define NA_OFI_OP_ERRORED   (1 << 4)
 
+/* The timeout (second) when give up for FI_EAGAIN's retry */
+#define NA_OFI_OP_RETRY_TIMEOUT ((double)90.0)
+
 /* Private data access */
 #define NA_OFI_CLASS(na_class)                                                 \
     ((struct na_ofi_class *) ((na_class)->plugin_class))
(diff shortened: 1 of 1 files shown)


soumagne:
@liuxuezhao I'm a little confused by this patch, though I understand your intent. If you get an error from the CQ, that means that the operation was successfully submitted in the first place, hence already removed from the retry queue so would not abort it but only complete with an error code there. Getting back to aborting retries I think that makes sense and we should do, probably based on a timeout of some sort though. When we retry and OFI returns EAGAIN for example the operation is not submitted yet so any error that occurs there would not be returned from the OFI CQ.

liuxuezhao:
> If you get an error from the CQ, that means that the operation was successfully submitted in the first place, hence already removed from the retry queue so you would not be able to abort it but it would only complete with an error code there.

Yes, the intent is at that time, it checks the retry op queue and abort all other OPs to the same target fi_addr. Because if based on current ofi rxm provider, those retry will always get EAGAIN so cannot get completion.

By some other test, this mercury patch cannot fully resolve the problem, possibly due to it cannot abort the new submitted OPs. We probably still need OFI change, or adding timeout tracking for each OP submitted to OFI that is more complex. Need some more discussions later.
Thanks.

liuxuezhao:
@soumagne anything else I need to for landing this patch? not very sure if the CI failure is related with this patch, thanks.

soumagne:
I simplified things a bit in #553. Have a look, it should be good to go now. Thanks!

PR #550, opened 2022-02-17
NA OFI: initialization and address management improvements

soumagne:


mathstuf:
Just tested this myself; with this patch a Linux client can successfully communicate with a macOS-hosted server (over `sockets://`).

soumagne:
thanks for confirming @mathstuf

PR #553, opened 2022-02-24
NA OFI: improve handling of retries after FI_EAGAIN return code

soumagne:
Abort retried ops after default 90s timeout

Abort ops to a target being retried after first NA_HOSTUNREACH error in CQ

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -252,6 +252,9 @@ static unsigned long const na_ofi_prov_flags[] = {NA_OFI_PROV_TYPES};
 #define NA_OFI_OP_QUEUED    (1 << 3)
 #define NA_OFI_OP_ERRORED   (1 << 4)
 
+/* Timeout (ms) until we give up on retry */
+#define NA_OFI_OP_RETRY_TIMEOUT (90 * 1000)
+
 /* Private data access */
 #define NA_OFI_CLASS(x)   ((struct na_ofi_class *) ((x)->plugin_class))
 #define NA_OFI_CONTEXT(x) ((struct na_ofi_context *) ((x)->plugin_context))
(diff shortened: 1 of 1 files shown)


src/na/na_ofi.c
```diff
@@ -4342,6 +4362,19 @@ na_ofi_cq_process_retries(struct na_ofi_context *na_ofi_context)
             }
             continue;
         } else if (rc == -FI_EAGAIN) {
+            hg_time_t now;
+
+            /* Do not retry past deadline */
+            hg_time_get_current_ms(&now);
+            if (hg_time_less(na_ofi_op_id->retry_deadline, now)) {
```
liuxuezhao, line 4369: Your change looks good to me, only one possible consideration is here my patch originally only check for RMA OP because that is what we really needed. 
The normal send/recv was excluded in my patch just for safe and avoid to introduce any possible trouble. My consideration is that for example for send OP (when send RPC request), if lots of RPC request send to a remote target (full filled OFI internal queue) that possibly cause EAGAIN returned from ofi, and if target server is in busy in relative long time (so don't recv the RPC in time) then it possibly cause EAGAIN returned to client-side in 90S. This case not only depend on network but also depend on server-side software handling for the RPCs.
But RMA is relative simpler that only depend on network (commonly wit hardware offloading), so I only checked for RMA OP.

But I cannot say checking all OPs is not good, probably fine to go with your change, just explain a little bit with that case.

soumagne: Yeah I do realize that, I feel like it would be awkward to have it only for one type of operations. I've been thinking about it and I think it should not cause any issues since it only kicks in when you get into the retry code path (which should really almost never happen except for rxm connections). And still assuming there's been a long time elapsed between the first FI_EAGAIN when you issue the op and the second time we retry when we enter `process_retries`, we only check the timeout _after_ we've got a second FI_EAGAIN so if by that time things have not resolved it's probably safe to assume that's an error and throw an NA_TIMEOUT back to the user. I think both cancelation from user and NA_TIMEOUT can also be used together, ie. one does not prevent the other. Anyway let's revisit if you see any issues but I'd rather have a consistent behavior across the board.

liuxuezhao: ok, sounds good to me, thanks.

PR #617, opened 2022-08-26
DAOS-11467 mercury: Add local address to ep when single IB device.

jgmoore-or:
With UCX, add local address to ep when single IB device is enabled.

Signed-off-by: Joseph Moore <joseph.moore@intel.com>


soumagne:
Now handled in #631

PR #623, opened 2022-09-07
NA OFI: Use strstr instead of strtok_r to parse URI

oserres:
Instead of using the strtok_r saveptr argument which is suppose to be opaque, use strstr to find the location of :. This makes the code slightly simpler and removes some warning brought up by some static code analysis tools.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -3994,13 +3994,10 @@ na_ofi_get_uri(const struct na_ofi_fabric *na_ofi_fabric,
     NA_LOG_SUBSYS_DEBUG(addr, "fi_av_straddr() returned %s", fi_addr_str);
 
     /* Remove unnecessary "://" prefix from string if present */
-    if (strstr(fi_addr_str, "://")) {
-        strtok_r(fi_addr_str, ":", &fi_addr_str_ptr);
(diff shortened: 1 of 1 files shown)


src/na/na_ofi.c
```diff
@@ -3995,7 +3995,7 @@ na_ofi_get_uri(const struct na_ofi_fabric *na_ofi_fabric,
 
     /* Remove unnecessary "://" prefix from string if present */
     if (strstr(fi_addr_str, "://")) {
-        strtok_r(fi_addr_str, ":", &fi_addr_str_ptr);
+        fi_addr_str_ptr = strstr(fi_addr_str, ":") + 1;
         rc = strncmp(fi_addr_str_ptr, "//", 2);
```
soumagne, line 3999: that's a good point and we could simplify the code even more actually since we don't need a second strstr() and the following strncmp, could you please do that instead ?

soumagne (changes requested):

src/na/na_ofi.c
oserres: Sure, I simplified the code further.

soumagne: Thanks!

soumagne (approved):

PR #625, opened 2022-09-08
HG: Avoid getting pointer of field if struct pointer is NULL

oserres:
The previous code performs the following operation even if hg_init_info is NULL: &hg_init_info->na_init_info.

This works correctly as na_init_info is the first field of struct hg_init_info.

This change explicitly tests for NULL pointer. Thus, making the code more explicit and simpler to understand (not relying on the order of the fields in the struct). It also avoids the issue being flagged by some code analysis tools.


soumagne (approved):
Looks good, thanks!

PR #639, opened 2023-01-22
HG util: Fix race condition in hg_mem_get_[huge]page_size

oserres:
Multiple threads access the page_size variable without synchronization. Fix the issue by using atomics.

Also remove the usage of ATOMIC_VAR_INIT on recent compilers as it is now deprecated and unnecessary.


soumagne (approved):
Thanks for the patch @oserres ! That looks good to me.

PR #655, opened 2023-02-24
Remove HG_ATOMIC_VAR_INIT

MaskRay:
ATOMIC_VAR_INIT has a trivial definition
`#define ATOMIC_VAR_INIT(value) (value)`,
is deprecated in C17/C++20, and will be removed in newer standards in newer GCC/Clang (e.g. https://reviews.llvm.org/D144196). Just remove the HG_ATOMIC_VAR_INIT wrapper.


soumagne (approved):
OK I'm definitely not opposed to simplification, thanks for submitting that PR!

PR #746, opened 2024-07-11
HG Core: add copy_on_multi_recv and multi_recv_copy_threshold

soumagne:
Use these two new parameters to fallback to memcpy to prevent starvation of multi-recv buffers and potential deadlock situations.

### DIFF
--- a/src/mercury_core_types.h
+++ b/src/mercury_core_types.h
@@ -110,6 +110,13 @@ struct hg_init_info {
      * existing buffers from being reposted.
      * Default value is: 4 */
     unsigned int multi_recv_op_max;
+
+    /* Controls when we should start copying data in an effort to release
+     * multi-recv buffers. Copy will occur when at most
+     * multi_recv_copy_threshold buffers remain. Value should not exceed
+     * multi_recv_op_max.
+     * Default value is: 0 (never copy) */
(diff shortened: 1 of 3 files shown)


src/mercury_core_types.h
```diff
@@ -110,6 +110,20 @@ struct hg_init_info {
...
      * Default value is: 4 */
     unsigned int multi_recv_op_max;
+
+    /* Force copy when receiving RPCs and multi-recv is being used (in an effort
+     * to release multi-recv buffers as soon as possible). This is particularly
+     * useful in large scale situations where multi-recv buffers tend to be
+     * quickly exhausted.
+     * Default value is: false */
+    bool copy_on_multi_recv;
+
```
frostedcmos, line 120: Looks like the intention for this experimental patch was to always copy, but since this is false by default it will not

soumagne: What made you think that? No the intent was to not copy by default. It's not really experimental I just made a patch that I can merge eventually. I also updated one of my cart PRs to set this to true to always copy as an experiment.

frostedcmos: Ah ok, the ticket PR for Saurabh to try only mentioned this PR and not the cart one, so i thought this was supposed to enable it by default. Makes sense now:)

PR #782, opened 2025-02-11
enable diagnostic counters without debug

mjmac:
In some cases, it may be useful to enable the counters without
enabling the full debugging configuration.

Signed-off-by: Michael MacDonald <mjmac@google.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,9 +175,22 @@ endif()
 option(MERCURY_ENABLE_DEBUG "Enable debug statements." OFF)
 if(MERCURY_ENABLE_DEBUG)
   set(HG_HAS_DEBUG 1)
+  # Always enable counters if debug is turned ON
+  set(MERCURY_ENABLE_COUNTERS "ON" CACHE BOOL "Enable diagnostics counters (enabled with debug)." FORCE)
 else()
   set(HG_HAS_DEBUG 0)
 endif()
+
+#------------------------------------------------------------------------------
+# Enable diagnostics counters separately from debug.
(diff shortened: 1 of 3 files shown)


mjmac:

mjmac:
@soumagne: Are these build failures expected? Glancing through them, it seems like an environmental issue (can't find MPI compiler?).

CMakeLists.txt
```diff
@@ -175,10 +175,22 @@ endif()
 option(MERCURY_ENABLE_DEBUG "Enable debug statements." OFF)
 if(MERCURY_ENABLE_DEBUG)
   set(HG_HAS_DEBUG 1)
+  set(HG_HAS_DIAG 1)
```
soumagne, line 178: I think here you should do instead (so that there is no conflict with the value of `HG_HAS_DIAG`):
```
  # Always enable counters if debug is turned ON
  set(MERCURY_ENABLE_COUNTERS "ON" CACHE BOOL "Enable diagnostics counters (enabled with debug)." FORCE)
```

soumagne (changes requested):

soumagne:
> @soumagne: Are these build failures expected? Glancing through them, it seems like an environmental issue (can't find MPI compiler?).

Yes, please rebase on top of latest master to pickup CI fixes. Also if you don't mind please add the prefix `HG core: ` to your commit. Thanks!

soumagne:
(oh and last thing, because you've signed the CLA, you don't need to sign off the commit)

soumagne (approved):
looks good, thanks!

PR #787, opened 2025-04-10
NA OFI: pass down na flags for firewall support in prov/tcp

jxiong:
This PR allows to pass down a flag to indicate if client bulk address is
behind firewall by using address deserialization callback functions.

This flag will lead to FI_FIREWALL_ADDR flag to be set in OFI.


src/mercury_bulk.h
```diff
@@ -18,12 +18,6 @@
 /* Public Macros */
 /*****************/
 
-/* The memory attributes associated with the bulk handle
- * can be defined as read only, write only or read-write */
-#define HG_BULK_READ_ONLY  (1 << 0)
-#define HG_BULK_WRITE_ONLY (1 << 1)
-#define HG_BULK_READWRITE  (HG_BULK_READ_ONLY | HG_BULK_WRITE_ONLY)
-
```
soumagne, line 26: please keep those like that, we need to only have those public, the rest must remain private.

src/mercury_bulk.h
```diff
@@ -200,7 +194,7 @@ HG_Bulk_get_segment_count(hg_bulk_t handle);
  *
  * \return Non-negative value
  */
-static HG_INLINE uint8_t
+static HG_INLINE uint32_t
```
soumagne, line 197: can you please keep it as uint8_t and fix HG_Bulk_get_flags() to instead return `((struct hg_bulk_desc_info *) handle)->flags & 0x3`

src/mercury_bulk.h
```diff
@@ -213,7 +207,7 @@ HG_Bulk_get_flags(hg_bulk_t handle);
  * \return Non-negative value
  */
 HG_PUBLIC hg_size_t
-HG_Bulk_get_serialize_size(hg_bulk_t handle, unsigned long flags);
+HG_Bulk_get_serialize_size(hg_bulk_t handle, uint32_t flags);
```
soumagne, line 210: keep unsigned long here

src/mercury_bulk.h
```diff
@@ -228,7 +222,7 @@ HG_Bulk_get_serialize_size(hg_bulk_t handle, unsigned long flags);
  */
 HG_PUBLIC hg_return_t
 HG_Bulk_serialize(
-    void *buf, hg_size_t buf_size, unsigned long flags, hg_bulk_t handle);
+    void *buf, hg_size_t buf_size, uint32_t flags, hg_bulk_t handle);
```
soumagne, line 225: same as above

src/mercury_bulk.h
```diff
@@ -344,7 +338,26 @@ HG_Bulk_cancel(hg_op_id_t op_id);
 struct hg_bulk_desc_info {
     hg_size_t len;          /* Size of region */
     uint32_t segment_count; /* Segment count */
-    uint8_t flags;          /* Flags of operation access */
+    uint32_t flags;         /* Flags of operation access */
+};
```
soumagne, line 342: it's good to keep `uint32_t` here.

src/mercury_bulk.h
```diff
@@ -362,7 +375,7 @@ HG_Bulk_get_segment_count(hg_bulk_t handle)
 }
 
 /*---------------------------------------------------------------------------*/
-static HG_INLINE uint8_t
+static HG_INLINE uint32_t
```
soumagne, line 378: see above

soumagne (changes requested):

soumagne (approved):

PR #799, opened 2025-06-06
NA BMI: Do not BMI_initialize() servers with address 0.0.0.0

chuckcranor:
The bmi code identifies servers by the hash of their listening address string.  If the listening address string on a set of servers running bmi+tcp is empty or 0.0.0.0 (INADDR_ANY), then it can hash to the same value causing hash conflicts.

To prevent listening address hash conflicts, we should not call BMI_initialize() with "tcp://0.0.0.0" as the listening address. Instead we use na_ip_pref_addr() to select a server IP, honoring na_init_info->ip_subnet (if provided).

Note that if the port number is zero, then BMI will let the kernel
choose the port with the bind() system call.   It needs to read
the selected port using getsockname() and append that value to
the listening address string to generate the correct hash of
the listening address.

Also fix the error recovery in na_bmi_initialize().   The
na_class->plugin_class is not set to na_bmi_class until the
very end of the function, so we should not key on na_class->plugin_class
to determine if we should free() or not.  Instead, we should
use na_bmi_class.

### DIFF
--- a/src/na/na_bmi.c
+++ b/src/na/na_bmi.c
@@ -1229,14 +1229,12 @@ na_bmi_initialize(
     struct na_bmi_class *na_bmi_class = NULL;
     char method_list[NA_BMI_ADDR_NAME_MAX] = {'\0'},
          listen_addr[NA_BMI_ADDR_NAME_MAX] = {'\0'},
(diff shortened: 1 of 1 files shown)


chuckcranor:
I sent this to @carns earlier to check.

soumagne (approved):

PR #820, opened 2025-09-04
NA UCX: flush endpoints before closing them

soumagne:
switch to using ucp_ep_close_nbx()

fixes to connect and address insertion

format fixes

### DIFF
--- a/src/na/na_ucx.c
+++ b/src/na/na_ucx.c
@@ -3099,6 +3099,21 @@ na_ucx_rma(struct na_ucx_class NA_UNUSED *na_ucx_class, na_context_t *context,
         ret, NA_BUSY, "Attempting to use OP ID that was not completed (%s)",
         na_cb_type_to_string(na_ucx_op_id->completion_data.callback_info.type));
 
+    /* There is no need to have a fully resolved address to start an RMA.
+     * This is only necessary for two-sided communication. */
+    /* The above assumption is now in question, so the following will resolve
+     * the address if required. */
+
+    /* Check addr to ensure the EP for that addr is still valid */
+    if (!(hg_atomic_get32(&na_ucx_addr->status) & NA_UCX_ADDR_RESOLVED)) {
+        ret = na_ucx_addr_map_update(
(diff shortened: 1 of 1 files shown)


src/na/na_ucx.c
```diff
@@ -1695,12 +1713,16 @@ na_ucp_listener_conn_cb(ucp_conn_request_h conn_request, void *arg)
...
         .addr = (const struct sockaddr *) &conn_request_attrs.client_address,
         .addrlen = sizeof(conn_request_attrs.client_address)};
     na_ucx_addr = na_ucx_addr_map_lookup(&na_ucx_class->addr_map, &addr_key);
-    NA_CHECK_SUBSYS_ERROR_NORET(addr, na_ucx_addr != NULL, error,
-        "An entry is already present for this address");
+
+    if (na_ucx_addr != NULL) {
+        NA_LOG_SUBSYS_WARNING(
+            addr, "An entry is already present for this address");
+        na_ucx_addr_release(na_ucx_addr, true, NULL);
```
gnailzenh, line 1724: I think I asked on the wrong PR yesterday, Joe answered me but I still don't quite understand. 
Is it ok to release the address without checking refcount? What if this address and associated EP are still using by someone else, will it cause memory corruption or other issue?

soumagne: it should be ok, we don't have much other choice I think, this fix was made by Joe for another issue and I just brought it up as part of this PR at the same time. We could add an extra warning there for sanity checks. If we end up in this code path this means that for this specific source address a new endpoint is established by UCX since this is triggered by the listener and afaik that means that the previous endpoint is no longer usable. Release will then force close the previous EP. I also don't see anything there that would cause a corruption since memory is not released.

soumagne: After spending some more time on it, I think the wise thing to do in this case would be to defer the connection request and let it complete (calling accept on it) only once the previous address has been fully released. As I said I'll remove that code from that PR and make a separate PR since this is unrelated to the original issue.

soumagne: Reworked this bit in https://github.com/mercury-hpc/mercury/pull/821

PR #852, opened 2026-02-27
Make it possible to specify multiple paths in NA_PLUGIN_PATH

mdorier:
Rational: I need to be able to test custom, dynamically-loaded plugins, however I also want to be able to use plugins installed in a default location (and do so in the same program, as Mochi is now able to run on top of multiple protocols at the same time).

The PR allows `NA_PLUGIN_PATH` to be a list of colon-separated paths. The special `"<default>"` string can be used still have NA look for plugins where it would normally have looked by default (i.e. in the same directory as where libna.so is installed), so for example we can do `NA_PLUGIN_PATH=/path/to/my/custom/plugins:<default>`.

The PR also makes cmake install the na_error.h and na_plugin.h headers if compiling with dynamic plugins.

### DIFF
--- a/src/na/CMakeLists.txt
+++ b/src/na/CMakeLists.txt
@@ -432,13 +432,25 @@ if(NA_HAS_SM)
   )
 endif()
 
+if(NA_HAS_DYNAMIC_PLUGINS)
+  set(NA_PUBLIC_HEADERS
+    ${NA_PUBLIC_HEADERS}
+    ${CMAKE_CURRENT_SOURCE_DIR}/na_plugin.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/na_error.h
+  )
+endif()
+
 #------------------------------------------------------------------------------
 # Specify project private header files
 #------------------------------------------------------------------------------
-set(NA_PRIVATE_HEADERS
-  ${CMAKE_CURRENT_SOURCE_DIR}/na_error.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/na_plugin.h
-)
+if(NA_HAS_DYNAMIC_PLUGINS)
+  set(NA_PRIVATE_HEADERS)
+else()
+  set(NA_PRIVATE_HEADERS
+    ${CMAKE_CURRENT_SOURCE_DIR}/na_error.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/na_plugin.h
+  )
+endif()
 
 if(NOT WIN32)
   set(NA_PRIVATE_HEADERS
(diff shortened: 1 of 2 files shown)


mdorier:

src/na/na.c
```diff
@@ -271,24 +271,90 @@ na_finalize(void) NA_DESTRUCTOR;
...
-        "if path indicated is not valid.",
-        plugin_path);
+        ret = na_plugin_scan_path(resolved_path, &na_plugin_dynamic_g);
+        NA_CHECK_SUBSYS_NA_FATAL(cls, done, ret,
+            "No usable plugin found in path (%s), consider setting "
+            "NA_PLUGIN_PATH if path indicated is not valid.",
+            resolved_path);
+    } else {
+        /* NA_PLUGIN_PATH set: scan each colon-separated path */
+        char *path_copy = strdup(plugin_path_env);
```
soumagne: there should be a null check here

src/na/na.c
```diff
@@ -271,24 +271,90 @@ na_finalize(void) NA_DESTRUCTOR;
...
+        free(path_copy);
+
+        if (total_count == 0) {
+            free(combined);
+            NA_LOG_SUBSYS_FATAL(cls,
+                "No usable plugin found in NA_PLUGIN_PATH (%s)",
+                plugin_path_env);
+        } else {
+            na_plugin_dynamic_g = combined;
+        }
```
soumagne: can you please move this entire block that you added into a separate routine ? I think that would just make things cleaner.

src/na/CMakeLists.txt
```diff
@@ -432,13 +432,25 @@ if(NA_HAS_SM)
   )
 endif()
 
+if(NA_HAS_DYNAMIC_PLUGINS)
+  set(NA_PUBLIC_HEADERS
+    ${NA_PUBLIC_HEADERS}
+    ${CMAKE_CURRENT_SOURCE_DIR}/na_plugin.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/na_error.h
+  )
```
soumagne, line 440: so you need that to build plugins externally right ?

soumagne (changes requested):

soumagne:
Also once you're done with the changes, please squash your commits and start your commit message with `NA: <description>`. Thanks!

src/na/na.c
mdorier: It's a few lines after; I'll move it.

src/na/CMakeLists.txt
mdorier: Yes

src/na/na.c
```diff
@@ -300,6 +307,76 @@ na_finalize(void)
 {
     na_plugin_close_all(na_plugin_dynamic_g);
 }
+
+/*---------------------------------------------------------------------------*/
+static void
+na_plugins_parse_path_list(const char *path_list)
```
soumagne: one minor change, please remove the s from na_plugins :)

soumagne (changes requested):

src/na/na.c
mdorier: Done

mdorier:
I think I have resolved everything? Is there anything else or can I squash the commits?

soumagne:
> I think I have resolved everything? Is there anything else or can I squash the commits?

Yes that's what I had mentioned above, once you're done with the changes, please squash your commits and start your commit message with `NA: <description>`. Then I will merge it.  Thanks!

soumagne:
Somewhat related to that PR, since you are starting to enable external NA plugins, we should think of a way of verifying the plugin version when it gets loaded. We don't really have anything that does that currently.

PR #857, opened 2026-05-29
NA UCX: flush endpoint after put

soumagne:
This ensures that data is correctly delivered upon completion of RMA puts and that the remote buffer can be safely re-used.


gnailzenh:
do you think it could have any potential performance impact? Though based on Shilong's test he ran, it looks fine.

PR #860, opened 2026-07-23
na/ofi: expand CXI NIC field to 32 bits to accommodate virtual functions

jordi-golarons:
From the perspective of this layer, virtual functions are indistinguishable from physical NICs — both are addressed by a NIC number. The CXI driver internally encodes the VF index into the upper bits of the NIC value, but this layer treats it as an opaque number and should not need to know about that encoding.

The previous 20-bit (C_DFA_NIC_BITS) NIC field silently truncated those upper bits. Expand struct na_ofi_cxi_addr to use uint64_t bitfields with a 32-bit NIC field, update the address parser to read a 64-bit hex value (SCNx64) and extract pid and nic via explicit bit masking, and fix na_ofi_cxi_to_key() to reconstruct the address from named fields instead of casting the struct pointer.

### DIFF
--- a/src/na/na_ofi.c
+++ b/src/na/na_ofi.c
@@ -521,11 +521,17 @@ struct na_ofi_cxi_addr {
     uint32_t nic : C_DFA_NIC_BITS;
     /* ignore other bits */
 };
-#else
+#elif FI_VERSION_LT(FI_COMPILE_VERSION, FI_VERSION(2, 7))
 struct na_ofi_cxi_addr {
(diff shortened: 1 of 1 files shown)


codambro:
libfabric changes were merged here:
https://github.com/ofiwg/libfabric/commit/932732e11157ab1272fa39857e44e0286affe842

The code was revised to no longer change the CXI address struct, so mercury should "just work" and this PR no longer needed

Issue #619, opened 2022-08-31
Deregistering in RPC handler


mdorier:
If an RPC handler calls `HG_Deregister` to deregister itself, a subsequent `HG_Respond` will cause a segfault, with the following backtrace:

```
#0  hg_set_struct (hg_handle=hg_handle@entry=0x55fe46cef810, hg_proc_info=0x594925d9ff758133, op=op@entry=HG_OUTPUT,
    struct_ptr=struct_ptr@entry=0x7f94fce16ef0, payload_size=payload_size@entry=0x7f94fce16ec0, more_data=more_data@entry=0x7f94fce16ebf "")
    at /tmp/mdorier/spack-stage/spack-stage-mercury-2.2.0-ftf2olh6pdj2dt2vn4fhcxifrucwuuba/spack-src/src/mercury.c:577
#1  0x00007f94fe9e2287 in HG_Respond (handle=handle@entry=0x55fe46cef810, callback=callback@entry=0x7f94fea3e600 <margo_cb>,
    arg=arg@entry=0x7f94fce16f50, out_struct=out_struct@entry=0x7f94fce16ef0)
    at /tmp/mdorier/spack-stage/spack-stage-mercury-2.2.0-ftf2olh6pdj2dt2vn4fhcxifrucwuuba/spack-src/src/mercury.c:2044
```

Ideally, an error in `HG_Respond` should be returned.

I haven't checked if other functions (like `HG_Get_input/output`) succeed, fail, or crash, but maybe it would be good to ensure a well-defined behavior of functions that use a handle associated with an RPC that has been deregistered.

soumagne:
Handles cache RPC info on server to prevent extra lookups from RPC ID hash table, we'll need to either refcount entries in hash table or ensure that we cannot deregister an RPC until all RPCs under that ID have been executed.

soumagne:
Linked to #534

soumagne:
Duplicate of #534

Issue #534, opened 2021-12-16
HG: safe mechanism to deregister an RPC while handles for that RPC are still in use


carns:
**Is your feature request related to a problem? Please describe.**

Imagine a hypothetical scenario in which a service is periodically receiving a particular RPC type.  The service then begins to shut down (without coordinating with clients) and deregisters that RPC as part of the shut down process.

In this case, a the service could have already begun executing handlers for the RPC, and those handlers will continue to execute despite deregistration.  Margo includes a workaround for this that seems to cover most cases by simply checking whether the registered data associated with a given RPC is NULL or not when it is retrieved https://github.com/mochi-hpc/mochi-margo/pull/170. 

**Describe the solution you'd like**

It may be cleaner if Mercury had a way to avoid impacting existing handles on a given RPC ID when deregistering.  For example it could deny new RPCs on that ID immediately, but use reference counting to defer full deregistration until in-flight handles associated with the ID are all closed.  There are probably other solutions; that's just one option.

**Describe alternatives you've considered**

So far it seems like in-flight RPCs aren't particularly harmed unless they rely on registered data associated with the RPC, but we are still testing.

shanedsnyder:
The Margo fix that Phil mentions is only part of the solution, as it just applies to some Margo boiler-plate logic that runs before user RPC handler code. It looks like service RPC handlers themselves have to be careful not to assume they will be able to retrieve data registered with the RPC -- that's not a huge deal to add safety checks there, but it would be nice if Mercury could provide some stricter guarantees in terms of lifetime of registered data for RPC handlers that are already executing.

soumagne:
no this has not been implemented yet

mdorier:
Bumping as I stumbled upon this problem once again today (and it took me a bit to figure it out). I guess I'm bound to rediscovering it once a year?... ah no, I skipped 2025 :-)

soumagne:
time flies :) I should take a look to see what we can do wrt reference counting

soumagne:
also maybe we should think about maintaining a top list of issues that you'd want to see fixed for mochi (in terms of priority) so that they don't get lost in the backlog...

PR #865, opened 2026-08-28
HG Core: ref-count RPC info to keep it alive across deregister

mdorier:
# Fix use-after-free of RPC info when an RPC is deregistered while in use

**Describe the problem**

Mercury freed an `hg_core_rpc_info` record as soon as you deregistered its RPC ID. The record had no reference count, so Mercury did not track the handles or the lookups that still pointed to it. When a request for the same RPC ID was already in the receive pipeline, or when another thread did a lookup at the same time, the record became a dangling pointer. A later read of the record caused a use-after-free and a crash.

Two failure modes occurred:

1. **Cached-pointer use-after-free.** The target side caches the record on a received handle (`hg_core_process()`) and the origin side caches it on the forward handle (`hg_core_set_rpc()`). A concurrent `HG_Core_deregister()` freed the record. The trigger step then read the freed record through the handle.

2. **Unlocked-lookup race.** `HG_Core_registered_data()` looked the record up, released the map read lock, and only then read `->data`. A concurrent `HG_Core_deregister()` freed the record between the lock release and the read. The function returned a garbage pointer instead of `NULL`, which defeated the `NULL` check that callers rely on.

This is a general defect. It happens in any service that deregisters an RPC while the Mercury instance keeps making progress and keeps receiving requests. Services that add and remove RPCs at run time hit it.

**Describe the fix**

Give `hg_core_rpc_info` the reference-count model that `hg_core_handle` and `hg_core_private_addr` already use. Keep the record alive while any handle or any lookup still points to it. Free the record only when the count reaches zero.

- Add an `hg_atomic_int32_t ref_count` field to `hg_core_rpc_info`.
- Set the count to `1` on insert for the map's own reference.
- `HG_Core_deregister()` removes the ID from the map and drops the map reference. The user free callback and `free()` now run from the reference release function, not from map removal.
- Take a reference each time a handle caches the record (target receive, origin forward). Release it when the handle stops pointing to the record (handle reset and handle free). The count stays balanced across the handle pool.
- Take the reference under the map read lock. Deregistration drops the map reference under the write lock, so the two cannot overlap. A caller either gets a valid reference, or gets `NULL` when the ID is already gone.
- Read `HG_Core_registered_data()` under the map read lock, so the record cannot be freed between the lookup and the read. A removed ID returns `NULL`.

After the fix, `HG_Core_deregister()` cannot free a record that a received handle still points to. A late trigger reads a valid record. A lookup for a removed ID returns `NULL`, so the existing caller checks work. Deregistration stays non-blocking.

**Compatibility**

- Public function signatures do not change. `HG_Core_deregister()` and `HG_Core_registered_data()` keep the same prototype.
- No behavior change for callers that do not deregister RPCs at run time.

**Affected files**

- `src/mercury_core.c`
- `src/mercury_core.h`

**Testing**

- Built with libfabric (`ofi+tcp`) and shared memory in a Debug configuration.
- `ctest` passes for all `ofi_tcp` and `sm` cases (20/20), unchanged from the baseline before the fix. The remaining `ofi_sockets` failures are unrelated: the `sockets` provider is not built into the libfabric in use.

### DIFF
--- a/src/mercury_core.c
+++ b/src/mercury_core.c
@@ -581,6 +581,12 @@ static HG_INLINE int
 hg_core_map_equal(hg_hash_table_key_t key1, hg_hash_table_key_t key2);
 
 /**
+ * Decrement reference count on RPC info and free it when it reaches zero.
+ */
+static void
+hg_core_rpc_info_decref(struct hg_core_rpc_info *hg_core_rpc_info);
+
+/**
  * Free value in map.
  */
 static void
@@ -591,6 +597,18 @@ hg_core_map_value_free(hg_hash_table_value_t value);
  */
 static HG_INLINE struct hg_core_rpc_info *
 hg_core_map_lookup(struct hg_core_map *hg_core_map, hg_id_t *id);
+
+/**
+ * Lookup entry for RPC ID and take a reference on it under the map lock.
+ */
+static HG_INLINE struct hg_core_rpc_info *
+hg_core_map_lookup_ref(struct hg_core_map *hg_core_map, hg_id_t *id);
+
+/**
+ * Lookup user data for RPC ID and read it under the map lock.
+ */
+static HG_INLINE void *
+hg_core_map_lookup_data(struct hg_core_map *hg_core_map, hg_id_t *id);
 
 /**
  * Insert new entry for RPC ID.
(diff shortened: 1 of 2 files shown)


mdorier:
Note: this fixes https://github.com/mercury-hpc/mercury/issues/619 and https://github.com/mercury-hpc/mercury/issues/534

soumagne (approved):
That looks good to me, thanks!

soumagne:
Merged in d1b90f8baac57f6e8c9762f70b3a4c723dba1614

PR #866, opened 2026-09-08
NA OFI: Only set NA_OFI_ALLOC_HUGE on successful hugepage allocs

codambro:
hg_mem_huge_alloc sets mem_ptr to MAP_FAILED, but subsequent check is only for NULL. Setting NA_OFI_ALLOC_HUGE incorrectly


codambro:

codambro:
i was wrong, theres a goto error setting it to NULL

